PR #1, opened 2022-08-16
Fix Github link in top right of page

Andy-Hay:


stanch:
Actually I changed this on purpose. I think having a link to Snowplow GitHub is more useful. On the pages we already have a link that says “edit this page on GitHub” which leads directly to the corresponding markdown file.

Andy-Hay:
Ah, okay :) 

Closing off the PR

PR #2, opened 2022-08-16
Added tracking

JOE-SNWPLW:


Andy-Hay (approved):
LGTM

stanch (changes requested):

snowplow.js
```diff
@@ -0,0 +1,97 @@
...
+  window.snowplow("enableActivityTracking:snplow5", 10, 10); // precise tracking for the unified log
+  
+  window.snowplow("trackPageView");
+}
+
+const setupBrowserTracker = () => {
+  if (ExecutionEnvironment.canUseDOM) {
+    let appId = 'test'
+    const host = window.location.hostname
+    if (host === 'docs.snowplowanalytics.com' || host === 'docs.snowplow.io') {
```
stanch, line 55: `snowplow.github.io` instead of `docs.snowplow.io` :)

snowplow.js
```diff
@@ -0,0 +1,97 @@
...
+      cookieDomain: ".snowplowanalytics.com",
+      cookieName: "_sp5_",
+      contexts: {
+        webPage: true,
+        performanceTiming: true,
+        gaCookies: true,
+      }
+    })
+
+    newTracker('gtm-eng', 'eng-gtm.snowplowanalytics.com', {
```
stanch, line 71: Let me check tomorrow what this is for...

snowplow.js
```diff
@@ -0,0 +1,97 @@
...
+  });
+  
+  window.snowplow("enableLinkClickTracking", true, true, null); 
+  
+  window.snowplow("enableActivityTracking:snplow5", 10, 10); // precise tracking for the unified log
+  
+  window.snowplow("trackPageView");
+}
+
+const setupBrowserTracker = () => {
```
stanch, line 51: So pretty!!!

snowplow.js
stanch: ...or is `docs.snowplow.io` something we are adding also?

JOE-SNWPLW: I thought as we will migrate from snowplowanalytics.com to snowplow.io that we would naturally move the docs there as well. Isn't that correct? 

Do you want to make the docs available on a GitHub url? I though we will add our own custom domain.

snowplow.js
JOE-SNWPLW: Sure, if it's obsolete I will remove it. This whole thing is so weird for me. We have GA, GTM, and 2 snowplow trackers all on the docs. It seems redundant but I don't know the rationale behind this setup..

snowplow.js
stanch: You are right, I only found out about the `snowplow.io` migration yesterday 🤦‍♂️ The GitHub URL will be hidden eventually, but I think it would be good to keep it for testing, until we do that.

snowplow.js
stanch: I confirmed that this can be removed.

snowplow.js
```diff
@@ -0,0 +1,97 @@
...
+  if (ExecutionEnvironment.canUseDOM) {
+    let appId = 'test'
+    const host = window.location.hostname
+    if (host === 'docs.snowplowanalytics.com' || host === 'docs.snowplow.io') {
+      appId = "docs2"
+    }
+
+    const snowplowTracker = newTracker('snplow5', 'collector.snowplowanalytics.com', {
```
stanch, line 59: This one we should replace with `collector-g.snowplowanalytics.com`

jbeemster: This is not going to ops1 pipeline is it?  I thought all web traffic was now going to our biz1 pipeline so it should instead be c.snowplow.io or c.snowplowanalytics.com ?

stanch (approved):

PR #5, opened 2022-08-22
wip: Make all internal links Markdown links

stanch:
Docusaurus can check for broken links if we spell them out as `.../index.md`.
Adding a script to turn all links into Markdown links and updating all the files.
Note that some links changed more than that because of redirects that were in place on the original site and are not in place now. (Even if we add those redirects later, it’s a good idea to keep links up to date with the file structure.)


stanch:
Not ready yet, there are 18 broken links we need to fix by hand. Only then the broken link checker can be enabled.

stanch:
OK, apparently links within raw HTML don’t work with `index.md` in them, so I will not change them (unless the link is broken and needs to be changed)

stanch:
Will reopen with an easier to review approach

PR #6, opened 2022-08-23
Make links checkable by Docusaurus

stanch:
Docusaurus can check for broken Markdown links if we spell them out as `.../index.md`.
This pull request will fix all currently broken links and make sure we have this check enabled going forward.


stanch:
Step 1: update the broken links with redirects

stanch:
Looks good, don’t want to torture anyone, so will self-approve.

PR #11, opened 2022-08-26
Cookie consent

JOE-SNWPLW:


src/pages/cookie-settings.mdx
```diff
@@ -0,0 +1,76 @@
...
+date: "2022-01-25"
+---
+
+import ManageCookieSettings from '@site/src/components/ManageCookieSettings'
+
+# Cookie settings
+
+### About this cookie policy
+
+This Cookie Policy explains what cookies are and how we use them. You should read this policy to understand what cookies are, how we use them, the types of cookies we use i.e, the information we collect using cookies and how that information is used and how to control the cookie preferences. For further information on how we use, store and keep your personal data secure, see our Privacy Policy.
```
stanch, line 12: Link to Privacy Policy (https://snowplowanalytics.com/privacy-policy/). I know it does not match 100% but it’s probably good enough

src/pages/cookie-settings.mdx
```diff
@@ -0,0 +1,76 @@
...
+
+You can at any time change or withdraw your consent from the Cookie Declaration on our website. Learn more about who we are, how you can contact us and how we process personal data in our Privacy Policy. Your consent applies to the following domains: docs.snowplowanalytics.com
```
stanch, line 14: on this website

make Privacy Policy a link

src/pages/cookie-settings.mdx
```diff
@@ -0,0 +1,76 @@
...
+
+:::note
```
stanch, line 16: I would move this down to “How can I control cookie preferences”

src/pages/cookie-settings.mdx
```diff
@@ -0,0 +1,76 @@
...
+      <td>The _sp5_ cookie, installed by Snowplow Analytics, stores information on how visitors use a website, while also creating an analytics report of the website's performance. Some of the data that are collected include the number of visitors, their source, and the pages they visit anonymously.</td>
+    </tr>
+  </tbody>
+</table>
+
+### How can I control the cookie preferences ?
+
+You can manage your cookies preferences by clicking on the "Settings" button and enabling or disabling the cookie categories on the popup according to your preferences.
+
+Should you decide to change your preferences later through your browsing session, you can click on the "Privacy & Cookie Policy" tab on your screen. This will display the consent notice again enabling you to change your preferences or withdraw your consent entirely.
```
stanch, line 74: This paragraph is pointless since we are already on this page, meaning the user has found it! Let’s remove.

cookieConsent.js
```diff
@@ -0,0 +1,54 @@
...
+  ],
+  "permissionLabels": {
+    "accept": "Accept",
+    "acceptAll": "Accept all",
+    "decline": "Decline"
+  },
+  "cookiePreferenceKey": COOKIE_PREF_KEY,
+  "header": {
+      "title": "cookie though?",
+      "subTitle": "You're probably fed up with these banners...",
```
stanch, line 29: Tiny nitpick, let’s use `’` instead of `'` since we are doing that below

stanch (changes requested):

src/pages/cookie-settings.mdx
```diff
@@ -0,0 +1,76 @@
...
+    <tr>
+      <td>_sp5_</td>
+      <td>The _sp5_ cookie, installed by Snowplow Analytics, stores information on how visitors use a website, while also creating an analytics report of the website's performance. Some of the data that are collected include the number of visitors, their source, and the pages they visit anonymously.</td>
+    </tr>
+  </tbody>
+</table>
+
+### How can I control the cookie preferences ?
+
+You can manage your cookies preferences by clicking on the "Settings" button and enabling or disabling the cookie categories on the popup according to your preferences.
```
stanch, line 72: either change this to “manage your consent”, or change the widget to say “settings” (which I think sounds more friendly)

src/components/ManageCookieSettings/index.js
```diff
@@ -0,0 +1,41 @@
...
+        setConsentState('Only the Necessary cookies are enabled')
+      }
+    }
+
+  }, [])
+
+  return (
+    <p>
+      <span className={styles.consentDesc}>Your cookie consent state:</span>
+      <span className={styles.consentState}>{consentState}</span>
```
stanch, line 37: I would remove bold and italic and “Your cookie consent state:”
Instead, just “You currently have ... enabled. [Settings button]”

src/pages/cookie-settings.mdx
JOE-SNWPLW: Is this 2 different things?
1. Change the last sentence to: "Your consent applies on this website" ?
2. Make the "Privacy Policy" a link similarly to the previous comment ?

src/pages/cookie-settings.mdx
JOE-SNWPLW: I think it makes sense to keep in "above the fold" (visible on screen by default) as it is in the current docs. Otherwise I think nobody would bother to scroll down to find it. (Assuming we want to optimize for people to use this - if that's not the case its a good idea to move it down)

src/components/ManageCookieSettings/index.js
JOE-SNWPLW: What would be the default case? (When someone opens this page from a direct link without ever consenting?) "You currently have no cookies enabled" ?

stanch: Yes

stanch (approved):

PR #15, opened 2022-08-29
<fix> broken table

pkutaj:
* See https://docs.snowplow.io/docs/pipeline-components-and-applications/stream-collector/configure#tls-port-binding-and-certificate-240


stanch (approved):

stanch:
Thanks!

PR #34, opened 2022-09-01
Fix Node.js v3 tracker initialization library and default protocol - Update for v3.6.0

igneel64:
### Description
Fix an error with the library name and change the default protocol. Should merge after https://github.com/snowplow/snowplow-javascript-tracker/pull/1101

### DIFF
--- a/docs/collecting-data/collecting-from-own-applications/javascript-trackers/node-js-tracker/node-js-tracker-v3/initialization/index.md
+++ b/docs/collecting-data/collecting-from-own-applications/javascript-trackers/node-js-tracker/node-js-tracker-v3/initialization/index.md
@@ -17,7 +17,7 @@ var tracker = snowplow.tracker;
 or, if using ES Modules, you can import the module like so:
 
 ```
-import { tracker, gotEmitter } from 'snowplow-tracker';
+import { tracker, gotEmitter } from '@snowplow/node-tracker';
 ```
 
 ### Configure Emitter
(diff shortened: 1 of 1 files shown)


stanch (approved):
LGTM, assuming the changes are correct :) I’ll defer validating that to others.

docs/collecting-data/collecting-from-own-applications/javascript-trackers/node-js-tracker/node-js-tracker-v3/initialization/index.md
```diff
@@ -31,7 +31,7 @@ A simple set up of this emitter might look like:
 ```
 const e = gotEmitter(
   'collector.mydomain.net', // Collector endpoint
-  snowplow.HttpProtocol.HTTPS, // Optionally specify a method - http is the default
+  snowplow.HttpProtocol.HTTPS, // Optionally specify a method - https is the default
```
matus-tomlein, line 34: Is there actually a default? It seems like it is a required property for the `gotEmitter` call and I don't see a place where the default is assigned, but I might just be confused.

matus-tomlein (approved):
LGTM! 👍 I don't understand how the defaults work in the NodeJS tracker but I trust you know better.

matus-tomlein:
Having made the comment above, I just noticed the PR for protocol defaults 🙈 So this is just a planned documentation, not one that should be merged right away, maybe we should set up a base branch like `release/js_tracker/3.6.0` where we could collect the issue documentation updates?

igneel64:
@matus-tomlein 
Thanks for the review! Yes sounds like a good idea for release updates.
Depending on what is included in the release group, I will make sure to create a release branch if we have more changes than this one.

PR #37, opened 2022-09-01
Add additional language styling

Andy-Hay:


docs/collecting-data/collecting-from-own-applications/google-amp-tracker/google-amp-1-0-0-2/index.md
```diff
@@ -12,8 +12,7 @@ To learn more about analytics for AMP pages see the [amp-analytics](https://www.
 
 Snowplow is natively integrated into the project, so pages optimized with AMP HTML can be tracked in Snowplow by adding the appropriate `amp-analytics` tag to your pages:
 
-```html
```
stanch, line 15: Both `html` and `xml` are listed as supported... https://prismjs.com/#supported-languages

docusaurus.config.js
```diff
@@ -93,19 +93,20 @@ const config = {
...
             label: 'Terms and conditions',
-            href: '/terms-and-conditions', 
+            href: '/terms-and-conditions',
           }
         ],
         copyright: `Copyright © ${new Date().getFullYear()} Snowplow Analytics Ltd. Built with Docusaurus.`,
       },
       prism: {
         theme: require('prism-react-renderer/themes/dracula'),
+        additionalLanguages: ['arduino', 'csharp', 'docker', 'gradle', 'java', 'php', 'properties', 'r', 'ruby', 'scala', 'swift'],
```
stanch, line 109: I think we should preemptively list all languages we use (`git grep -Eho '```[a-z]+' '*.md' | sort | uniq | sed 's/```//'`), so that we don’t depend on what Prism deems to be good defaults.

stanch (changes requested):

docusaurus.config.js
Andy-Hay: You can't - build fails if you add language X and X isn't supported by Prism.  Gives error like this for `html`

```
Error: Cannot find module './prism-html'
    at Array.forEach (<anonymous>)
[ERROR] Unable to build website for locale en.

● Client █████████████████████████ cache (99%) shutdown IdleFileCachePlugin
 resolve build dependencies

● Server █████████████████████████ cache (99%) shutdown IdleFileCachePlugin
 process pending cache items

[ERROR] Error: Failed to compile with errors.
    at /Users/andyhay/code/snowplow.github.io/node_modules/@docusaurus/core/lib/webpack/utils.js:180:24
    at /Users/andyhay/code/snowplow.github.io/node_modules/webpack/lib/MultiCompiler.js:554:14
    at processQueueWorker (/Users/andyhay/code/snowplow.github.io/node_modules/webpack/lib/MultiCompiler.js:491:6)
    at process.processTicksAndRejections (node:internal/process/task_queues:77:11)
[INFO] Docusaurus version: 2.0.1
```

stanch: That’s annoying :(

docs/collecting-data/collecting-from-own-applications/google-amp-tracker/google-amp-1-0-0-2/index.md
Andy-Hay: Moved to markup as html/xml seem to be aliases of same markup Prism configuration.  When trying to add them specifically they didn't exist.

I thought it better to move these to markup as thats what they are and not to add future confusion when html/xml aren't supported by Docusaurus specifically and can't be added as additional types.  It is confusing.  See error that's generated in comments below.

stanch: I would keep them as `html` / `xml` just because it’s more descriptive, and may be an advantage if we move to a different highlighter in the future.

docusaurus.config.js
stanch: In this case, keep it as you did, but add a comment explaining that we can’t add more (and a link to where the default ones are defined)

docs/collecting-data/collecting-from-own-applications/google-amp-tracker/google-amp-1-0-0-2/index.md
Andy-Hay: That's true - will drop out the html/xml changes

docusaurus.config.js
Andy-Hay: Reverted other changes, just single change in config file with comments referencing the links and warning about doing a build to check (yarn start just fails with no displayed error)

stanch (approved):

PR #41, opened 2022-09-01
Add external redirects for API references

paulboocock:
There are some API Reference websites currently running on other Github Pages and previously we redirected these from a nav bar item. I've tried to achieve the same here to maintain consistency, theres a quick page change before it redirects but hopefully this is ok for now. We could perhaps find a new way to create these links too but I wanted to get them fixed in the first instance.

I tried to do some other stuff with the sidebar but since we autogenerate it (as we should!) I couldn't work out a way to inject an external link in there.


paulboocock:
@stanch Updated to use a custom sidebar generator which switches the Doc Items out for Link Items where there is a `href` in the Front Matter of the page.
There's also no need for the `sidebarPath` as when undefined it defaults to autogenerated: https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-docs#sidebarPath
This means I can keep the custom sidebar logic in sidebars.js rather than cluttering docusaurus.config.js

sidebars.js
```diff
@@ -11,21 +11,37 @@
...
-  ],
-   */
+const swap = (items, itemsWithHref) => {
+  const result = items.map((item) => {
+    if (item.type === 'category') {
+      return { ...item, items: swap(item.items, itemsWithHref) }
+    }
+
+    if (itemsWithHref[item.id]) {
+      item = {
```
stanch, line 21: Any reason not to put the return here?

sidebars.js
```diff
@@ -11,21 +11,37 @@
...
+        type: 'link',
+        label: itemsWithHref[item.id].title,
```
stanch, line 23: Why not just `item.label`?

sidebars.js
```diff
@@ -11,21 +11,37 @@
...
+  })
+
+  return result
+}
+
+// Switch out doc items for external links where required
+const swapDocItemsToLinkItems = (items, docs) => {
+  const itemsWithHref = {}
+
+  for (const docItem of docs) {
```
stanch, line 38: I have a feeling that if we use `sidebar_custom_props`, this will not be necessary. From [the docs](https://docusaurus.io/docs/sidebar/items#sidebar-item-doc):
> You can use `sidebar_custom_props` to declare custom metadata for a doc page.

> Sidebar custom props is a useful way to propagate arbitrary doc metadata to the client side, so you can get additional information when using any doc-related hook that fetches a doc object.

> `customProps?: Record<string, unknown>`

sidebars.js
```diff
@@ -11,21 +11,37 @@
...
+  const result = items.map((item) => {
+    if (item.type === 'category') {
+      return { ...item, items: swap(item.items, itemsWithHref) }
+    }
+
+    if (itemsWithHref[item.id]) {
+      item = {
+        type: 'link',
+        label: itemsWithHref[item.id].title,
+        href: itemsWithHref[item.id].href,
```
stanch, line 24: `item.customProps.href`?

sidebars.js
paulboocock: Because `item` only has `id` and `type` here, as it's already been through the default sidebar generator.

I did ponder if I could edit the items _before_ running through the default generator but I wasn't sure if the default generator would support `link` types. I'll have a play, see if it makes the code cleaner (and if it works).

sidebars.js
paulboocock: Not really. I have to `return item` at the end either way, and I'm not a massive fan of returning things mid way through a function (either early return immediately or at the the end). I'm also not a fan of mutating a field like this but you can't win em all.

sidebars.js
paulboocock: So this didn't work. `customProps` are sent to the component but just become part of the `frontMatter` in this function, so no different to `href`. I'd have to access it with `docItem.frontMatter.sidebar_custom_props.href` which seems a bit pointless when I'm just trying to build `link` type.

They don't magically appear on the generated doc item, as they don't actually get mapped onto doc items: https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-plugin-content-docs/src/sidebars/generator.ts#L139-L152

What this did make me realise though, is that my custom item generator didn't consider all the properties that could be passed to the component.

The `SidebarItemLink` type and `SidebarItemBase` describe these
https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-plugin-content-docs/src/sidebars/types.ts#L43-L48

So I've made sure to map all the properties (except autoAddBaseUrl which seems to be some internal flag because it throws a validation exception if I try to map it).

sidebars.js
paulboocock: I need to run the default generator first, because it has no concept of `link` items in it, meaning it'd ignore any fiddling I did with the original arguments.

https://github.com/facebook/docusaurus/blob/main/packages/docusaurus-plugin-content-docs/src/sidebars/generator.ts

stanch (approved):

PR #42, opened 2022-09-01
Initial addition of clarification of AWS principles of GTM SS

Andy-Hay:


docs/forwarding-events-to-destinations/forwarding-events/google-tag-manager-server-side/index.md
```diff
@@ -25,18 +25,26 @@ GTM SS with Snowplow can be setup in two different configurations.
...
 
 Use GTM SS to relay raw events before the Snowplow pipeline to destinations, including to your Snowplow pipeline.
 
 | Open Source | Snowplow BDP |
-| --- | --- |
-| ✔️ | ✔️ |
+|:-----------:|:------------:|
+|     ✔️      |      ✔️      |
+
+### Principles for AWS deployment
```
jbeemster, line 41: This might be better as a general purpose section for BDP around how to co-locate infrastructure with what we deploy and to then link back to it from here?  This principal is true of many things:

1. Redshift Clusters
2. BI Tools
3. OpenSearch Clusters

All should be segmented and away from what we manage - wdyr?

Andy-Hay: I think that there should be a general all encompassing section around infrastructure that could be linked to at various points calling out the specifics for feature X or Y.  We all tend to go to the documentation section that we want to read and expect it to have all that we need.  Having both the top level principles and the specific function principles (if specifics need calling out) is best way to go.

I propose raising issue to get top level infrastructure principles for each product written up.  Use that as a link within all these optional/add-on features to re-state the principles.  Use the proposed feature specific change now (with possible wording changes) and add a link to the top level principles once available.  How does that sound ?  I know that structure of docs will be changing and the presence of the infrastructure principles can be a requirement of that.

paulboocock: Whichever way we go here, I think we need to make it clear this is a note for BDP Customers. Just to reduce any confusion for others who might be using Snowplow outside of BDP and be happy to co-locate their GTM SS and Snowplow.

Andy-Hay: Reworked the paragraph into an info block targeted at BDP customers.  More generalised principles would need adding as we improve the overall content and then can be referenced from a number of locations.  Does this seem better ?

image

paulboocock (approved):

PR #45, opened 2022-09-02
Add SnowplowDocCard / SnowplowDocCardList components for wrapped text on TOC pages

Andy-Hay:
Created two new components (cloned Docusaurus ones and prefixed with Snowplow) to provided wrapped title and description in the TOC pages.

Upgraded to Docusaurus v2.1.0.

Note that definition of description is buried quite deep in Docusaurus so not able to change it without a lot of surgery.  Definition is use description metadata and if not present use the first line of the content.  This is a fundamental of how the sidebars / tocs are based on so that minimal data is being carried around.  We do need to add better descriptions to drive better search/SEO results.  Some first lines are very poor indeed. At least it can all be seen as its wrapped.  Option is to not display the description until we have added better descriptions on TOC pages. 

image

### DIFF
--- a/docs/collecting-data/collecting-data-from-third-parties/index.md
+++ b/docs/collecting-data/collecting-data-from-third-parties/index.md
@@ -12,7 +12,6 @@ View our [set of trackers for tracking events from your own applications](/docs/
 
 ```mdx-code-block
 import SnowplowDocCardList from '@site/src/components/SnowplowDocCardList';
-import {useCurrentSidebarCategory} from '@docusaurus/theme-common';
 
-<SnowplowDocCardList items={useCurrentSidebarCategory().items}/>
+<SnowplowDocCardList />
 ```
--- a/docs/collecting-data/collecting-data-from-third-parties/index.md
+++ b/docs/collecting-data/collecting-data-from-third-parties/index.md
@@ -11,8 +11,8 @@ Webhooks allow this third-party software to send their own internal event stream
 View our [set of trackers for tracking events from your own applications](/docs/collecting-data/collecting-from-own-applications/index.md).
 
 ```mdx-code-block
-import DocCardList from '@theme/DocCardList';
+import SnowplowDocCardList from '@site/src/components/SnowplowDocCardList';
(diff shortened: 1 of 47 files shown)


Andy-Hay:
Need to rerun pipeline once we're on v2.1.0 as its been built on it (some corrections in original DocCard / DocCardList functionality mean that imports aren't present in v2.0.1 pipeline).

Andy-Hay:
This highlights that the TOC boxes are using the first line of text in the Markdown (after Frontmatter) as there aren't many Frontmatter `description`'s 😢

stanch:
Thanks @Andy-Hay! Any reason why the upgrade to 2.1 is not included in this PR? I think with that we would also be able to get rid of the `item` properties.

I’d also consider making the doc card list a global import. That way adding it would be minimal work, basically just `<SnowplowDocCardList/>`.

stanch:
P.S. I think we also abuse the card lists a bit. If a page is _just_ a card list, I’d delete `index.md` and replace it with an equivalent `_category_.yml` (https://docusaurus.io/docs/sidebar/autogenerated#category-item-metadata). The corresponding sidebar item then will cease being a page you can navigate to.

paulboocock:
> P.S. I think we also abuse the card lists a bit. If a page is _just_ a card list, I’d delete `index.md` and replace it with an equivalent `_category_.yml` (https://docusaurus.io/docs/sidebar/autogenerated#category-item-metadata). The corresponding sidebar item then will cease being a page you can navigate to.


The "listings" pages were often available on the old docs site, would removing these cards views lead to potentially broken urls in the wild?

Andy-Hay:
Upgrade added.

TOC pages need to stay as they are not just the autogenerated content.  Some have lead text which gives context to the TOC.  They are quite genuine pages that can/are linked to.  

Option to not display the wrapped description offered as we don't have many Frontmatter description attributes, so it defaults to first line of content (which can be a little raw).

stanch:
@joe-miskolczy could you take a look at the components/swizzling etc? I don’t feel qualified

PR #54, opened 2022-09-07
Clarify a note about what happens to identifiers when anonymous tracking is enabled in JS tracker

matus-tomlein:
This is a small change that expands on the note field in JS tracker docs to give more details on what happens when users call `enableAnonymousTracking`. Also made a few other notes use the `:::note` format.


stanch (approved):
The fact that these files are exactly the same, save for the code examples, makes me wonder if JS/Browser docs should be in the same place but with tabs...

matus-tomlein:
@stanch i was thinking the same, most of the differences are in the snippets and can be easily solved with tabs. Plan to merge it when there is a chance.

PR #60, opened 2022-09-07
Move `page/index.md` to `page.md`

stanch:
I ran the following:

```bash

for file in $(git ls-files '**/index.md'); do
    if [ $(ls $(dirname $file) | wc -l | tr -d ' ') = 1 ]; then
        git mv $file $(echo $file | sed s!/index.md!.md!)
    fi
done

for file in $(git grep -l -E '\.\./[^)]+/index\.md' '*.md'); do
    sed -E 's!\.\./([^)]+)/index\.md!\1.md!g' $file | sponge $file
done

for link in $(git grep -Eoh '[("]/docs[^)"#?]+' docs | cut -c2- | sort | uniq); do
    relative=$(echo $link | cut -c2-)
    if [ ! -f $relative ]; then
        replacement=$(echo $link | sed 's!/index.md!.md!')
        for file in $(git grep -rl $link docs); do
            sed -E 's!'$link'!'$replacement'!g' $file | sponge $file
        done
    fi
done
```


paulboocock:
I thought I wanted this, but now I'm not so sure. I think the lack of consistency this creates makes editing and educating others how to edit more complicated. Currently the folder/index.md structure is consistent and matches the url structure of the site. I'd propose we don't do this now, apologies!

PR #66, opened 2022-09-12
Bump ES loader examples to 2.0.7

istreeter:


stanch (commented):
Looks good, should we update this as well? https://docs.snowplow.io/docs/getting-started-on-snowplow-open-source/setup-snowplow-on-aws/setup-destinations/elastic/installing-the-elasticsearch-loader/ Still says `0.10.0`

stanch:
And also https://docs.snowplow.io/docs/pipeline-components-and-applications/version-compatibility-matrix/latest-version-compatibility-matrix/

istreeter:
Good points -- the page ["Installing the elasticsearch loader"](https://docs.snowplow.io/docs/getting-started-on-snowplow-open-source/setup-snowplow-on-aws/setup-destinations/elastic/installing-the-elasticsearch-loader/) is a complete mess:

- The assets it refers to don't exist any more! (http/tcp/tcp2x)
- Doesn't mention docker images.
- Why are we telling users to build anything from source?
- The page doesn't actually say how to install the loader, which is what the title promises.  How do you actually run it and where?
- Version number is wrong

In your opinion, what should the "install" page offer to readers?  Like, what is it's reason for existing, over and beyond the main elasticsearch loader documentation?

If it's intended as a stripped down version of the _real_ elasticsearch loader documentation, then we should delete it entirely.  Replace it with a link to the _real_ docs.

If it's intended to add some kind of user-friendly narrative of using snowplow for the first time....  then it needs re-writing entirely.  And maybe it should be replaced with "go use this terraform module" (which didn't exist when that page was first written).

istreeter:
Regarding the compatibility matrix.... this normally only gets updated whenever there is an umbrella release, which is 2 or 3 times a year.  It is a more conservative list of versions, which we know to be stable.

docs/getting-started-on-snowplow-open-source/setup-snowplow-on-aws/setup-destinations/elastic/index.md
```diff
@@ -5,3 +5,5 @@ sidebar_position: 0
 ---
 
 The Elasticsearch loader reads enriched data from the enriched Kinesis stream and streams it into Elasticsearch in near real-time.
+
+The [Elasticsearch loader setup guide](/docs/pipeline-components-and-applications/loaders-storage-targets/elasticsearch-loader/#setup-guide) has full instructions for running and configuring the loader.
```
stanch, line 9:
```suggestion
The [Elasticsearch loader setup guide](/docs/pipeline-components-and-applications/loaders-storage-targets/elasticsearch-loader/index.md#setup-guide) has full instructions for running and configuring the loader.
```

stanch: See https://github.com/snowplow/snowplow.github.io#links

stanch (approved):

PR #72, opened 2022-09-14
Remove listen from repeater launch

jbeemster:
When debugging this with an OS user discovered that the guide says to set listen as an input command for the repeater which the repeater does not actually understand.  Removing this line got it working...

Looking in the code I can see listen however as a valid command so I am not sure whats going on here to be honest.

https://github.com/snowplow-incubator/snowplow-bigquery-loader/blob/master/modules/repeater/src/main/scala/com.snowplowanalytics.snowplow.storage.bigquery.repeater/RepeaterCli.scala#L72

Hoping @istreeter can help demystify!

### DIFF
--- a/docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/index.md
+++ b/docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/index.md
@@ -279,7 +279,6 @@ We recommend constantly running Repeater on a small / cheap node or Docker conta
 $ docker run \
     -v /path/to/configs:/configs \
     snowplow/snowplow-bigquery-repeater:1.4.2 \
-    listen \
     --config=/configs/bigquery.hocon \
     --resolver=/configs/resolver.json \
     --bufferSize=20 \ # size of the batch to send to the dead-letter bucket


istreeter (approved):
This is the correct fix.  The documentation was wrong.

The repeater does not take a sub-command name as the first argument.  You're right [the line you linked to](https://github.com/snowplow-incubator/snowplow-bigquery-loader/blob/master/modules/repeater/src/main/scala/com.snowplowanalytics.snowplow.storage.bigquery.repeater/RepeaterCli.scala#L72) is called the `ListenCommand` but that's just a variable name.  The repeater only does one thing, and that is to listen.

This makes more sense if you compare it with [the mutator implementation](https://github.com/snowplow-incubator/snowplow-bigquery-loader/blob/master/modules/mutator/src/main/scala/com/snowplowanalytics/snowplow/storage/bigquery/mutator/MutatorCli.scala#L95-L106) which makes use of `Opts.subcommand` so it accepts a few different subcommands as the first argument.

stanch (approved):

stanch:
Interestingly, in docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/previous-versions/bigquery-loader-0-6-0/index.md only mutator had a listen

PR #79, opened 2022-09-15
Update Micro docs for 1.3.3 (close #78)

### LINKED ISSUE #78: Update Micro docs for 1.3.3

greg-el:
This PR updates docs that refer to previous versions of Micro to the latest `1.3.3` release

### DIFF
--- a/docs/managing-data-quality/testing-and-qa-workflows/set-up-automated-testing-with-snowplow-micro/index.md
+++ b/docs/managing-data-quality/testing-and-qa-workflows/set-up-automated-testing-with-snowplow-micro/index.md
@@ -4,6 +4,11 @@ date: "2021-03-26"
 sidebar_position: 1000
 ---
 
+```mdx-code-block
+import {versions} from '@site/src/componentVersions';
+import CodeBlock from '@theme/CodeBlock';
+```
+
 # Snowplow Micro
 
 [![Latest Docker image version](https://img.shields.io/docker/v/snowplow/snowplow-micro?sort=semver)](https://hub.docker.com/r/snowplow/snowplow-micro)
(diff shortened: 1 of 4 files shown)


docs/managing-data-quality/testing-and-qa-workflows/set-up-automated-testing-with-snowplow-micro/index.md
```diff
@@ -181,7 +186,7 @@ An example of a response with one event can be found below:
       "name_tracker":"sp1",
       "v_tracker":"js-3.1.3",
       "v_collector":"ssc-2.3.1-stdout$",
-      "v_etl":"snowplow-micro-1.3.1-common-2.0.2",
+      "v_etl":"snowplow-micro-${versions.snowplowMicro}-common-2.0.2",
```
stanch, line 189: will the `common-2.0.2` part change over time?

stanch (approved):
Looks good, just wondering if the part I pointed out above will be a problem in the future (i.e. we update the version in the `.js` file, but forget we also need to update that part.

docs/managing-data-quality/testing-and-qa-workflows/set-up-automated-testing-with-snowplow-micro/index.md
greg-el: I'm thinking now that it probably doesn't make a huge amount of sense to use the variable here. Since it's an example of an event that's been copied and pasted, bumping the micro version in the payload seems a bit strange to me now.

stanch: Good point! Overlooked that. I guess it should be reverted then.

stanch (approved):

PR #85, opened 2022-09-21
Add notes regarding activity tracking callback documentation in v3 JS tracker docs

matus-tomlein:
Adds a note that the schema used in the example when enabling activity tracking callback in JS tracker is invalid as it is often being copied by users which then results in their events not showing up among good events in the warehouse.


matus-tomlein:
@paulboocock @igneel64 Shouldn't we create a valid schema for this use case? I see users copying this sample code and then being confused that they are not receiving the events repeatedly. Event if the schema would not fit everyone, at least we could give a working example?

paulboocock:
Yeah, I actually think in v4 we should wrap this whole feature up so users don't need to callback. Everyone uses it for the same reason and if we get an official schema we can work it into the official data models.

paulboocock:
Figured it was worth getting this out anyway (since it looks much better! and has a helpful note for the state of things today). I do still ponder if there's a nicer way to do this, but we'd have to be mindful of users who need to "trigger" the activity event where they never have an unload event (SPA for example).

PR #93, opened 2022-09-26
Fix `track_unstruct_event()` typos

esho:
Fix `track_unstruct_event()` typos

### DIFF
--- a/docs/collecting-data/collecting-from-own-applications/python-tracker/tracking-specific-events/index.md
+++ b/docs/collecting-data/collecting-from-own-applications/python-tracker/tracking-specific-events/index.md
@@ -21,7 +21,7 @@ In addition, Snowplow has a wide selection of pre-defined events and associated 
 | [`track_ecommerce_transaction_item()`](#ecommerce-transaction-item) | Track an item of an ecommerce transaction |
 | [`track_add_to_cart()`](#add-to-cart) | Track an add to cart event |
 | [`track_remove_from_cart()`](#remove-from-cart) | Track a remove from cart event |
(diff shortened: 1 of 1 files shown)


paulboocock:
Thanks for the PR @esho 

Unfortunately, the change you've suggested is actually incorrect as `track_struct_event()` is one of the predefined built in functions, for tracking "classic" events with Category, Action, Value, etc. (i.e. very structed vs a self describing event).

This table is attempting to show the predefined events, that are available beyond `track_self_describing_event()`. However, I do feel like this table draws the eye more than the paragraph above it which describes `track_self_describing_event()`. If you'd still like to contribute, updating this table to add `track_self_describing_event()` to the top of this table and changing the line before it to something like "Snowplow has a wide selection of event types and associated methods for tracking:"

paulboocock:
I also now see that I think you were trying to get the table to link to the bottom of the page: https://docs.snowplow.io/docs/collecting-data/collecting-from-own-applications/python-tracker/tracking-specific-events/#truck_unstruct_event

However, I think it should actually link to here:
https://docs.snowplow.io/docs/collecting-data/collecting-from-own-applications/python-tracker/tracking-specific-events/#track-structured-events
which is about half way down. We try not to mention `track_unstruct_event` as its legacy terminology and only here because we haven't fully removed it from this tacker yet.

It looks like all the anchor links in this table are broken actually :-/ 

I'm happy to fix this but if you'd like to beat me too it then please go ahead!

stanch:
Fixed in #182, following @paulboocock’s suggestion.

PR #97, opened 2022-09-29
Add S3 Loader 2.2.3

spenes:


stanch:
A few things:
* [The upgrade guide](https://deploy-preview-97--snowplow-docs.netlify.app/docs/pipeline-components-and-applications/loaders-storage-targets/s3-loader/upgrade-guides/2-2-0-upgrade-guide/) this still has `2.0.0` examples, should we keep this up to date with the latest 2.2? (In which case maybe it’s not a bad idea to rename the var to `s3Loader22x` or something like that, so that we don’t forget about it later.)
* Unrelated, but there are a few links wrongly formatted as code on [this page](https://deploy-preview-97--snowplow-docs.netlify.app/docs/pipeline-components-and-applications/loaders-storage-targets/s3-loader/monitoring/), could you fix them?
* Also unrelated, could you groom the table [here](https://deploy-preview-97--snowplow-docs.netlify.app/docs/pipeline-components-and-applications/loaders-storage-targets/s3-loader/configuration-reference/)? Just add the header like in other config tables and code-fence the parameter names on the left.

spenes:
@stanch I've addressed your feedback. Could you give it one more look, please ?

stanch (approved):

PR #114, opened 2022-10-05
Don’t let external search engines index old versions

stanch:
This way we will stop old pages from appearing in external search results.

At the same time, I’ve set [`ignoreNoIndex`](https://www.algolia.com/doc/tools/crawler/apis/configuration/ignore-no-index/) to true via the Algolia UI, so that our own search can still surface those pages.

### DIFF
--- a/src/theme/MDXContent/index.js
+++ b/src/theme/MDXContent/index.js
@@ -1,6 +1,7 @@
 import React from 'react'
 import MDXContent from '@theme-original/MDXContent'
 import Admonition from '@theme-original/Admonition'
+import Head from '@docusaurus/Head'
 import { useLocation } from '@docusaurus/router'
 
 // yeah, we need to standardize things!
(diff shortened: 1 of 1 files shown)


stanch:
```
$ http https://deploy-preview-114--snowplow-docs.netlify.app/docs/collecting-data/collecting-from-own-applications/java-tracker/previous-versions/java-tracker-v0-12/ | pcregrep -o '<meta[^>]+robots[^>]+>'
<meta data-rh="true" name="robots" content="noindex, follow">

$ http https://deploy-preview-114--snowplow-docs.netlify.app/docs/collecting-data/collecting-from-own-applications/java-tracker | pcregrep -o '<meta[^>]+robots[^>]+>'

```

paulboocock (approved):
Nice. I'm cool with this! 

Even makes me wonder if we'd want algolia to stop indexing them too but I guess because version selecting is challenging here atm thats probably not a good idea. The only reason I really suggest it is the previous versions pages are often inconsistently versioned (sometimes its a page per patch, minor or just majors 🤷 ) so landing on latest all the time seems like a good bet from any search engine. Searching for "Thing vx.x.x" seems very unlikely to take you anywhere useful anyway but you might accidently end up on an old page from the algolia search.

PR #116, opened 2022-10-06
Bump Python Version

Jack-Keene:
Bump Python version to 0.11.0


stanch (changes requested):
It would be good to use the same approach we have for other components, where we declare the latest version in `src/componentVersions.js` and use the variable in the docs. This way, when the time comes to update the version, you only need to update it in a single place.

See #103 for an example commit that follows this approach. See also https://docusaurus.io/docs/markdown-features/react#markdown-and-jsx-interoperability for some issues you might face (e.g. when using the variable inside a table).

stanch (approved):

PR #117, opened 2022-10-07
Update Flutter tracker docs for v0.2.0 (close #82)

mscwilson:
Updates the docs for the new version, and corrects a mistake (trackers can't be made in `initState()`, because it can't be async).

I noticed that the code blocks are labelled cpp. Do/can we have Dart syntax highlighting enabled?

PR #118, opened 2022-10-07
Add Dart to Prism languages

mscwilson:


mscwilson:
This change would be better as part of #117 !

PR #130, opened 2022-10-12
Fix RDB Loader config reference table format

Andy-Hay:
Fixes bad formatting in two sections of tables for Snowflake and Databricks

- https://docs.snowplow.io/docs/pipeline-components-and-applications/loaders-storage-targets/snowplow-rdb-loader/loading-transformed-data/rdb-loader-configuration-reference/#snowflake-loader-storage-section
- https://docs.snowplow.io/docs/pipeline-components-and-applications/loaders-storage-targets/snowplow-rdb-loader/loading-transformed-data/rdb-loader-configuration-reference/#databricks-loader-storage-section


stanch:
Thanks! Could you also add table headers (Parameter / Description) and fix these?
image
image

Andy-Hay:
Removed the overticking and added the table headings as suggested.

PR #161, opened 2022-10-28
Update tracker configuration with a new cookie name for the biz1 tracker

matus-tomlein:
Updates the configuration for `snplow5` and `biz1` trackers:

* sets the method to POST on both trackers
* sets `cookieSameSite` to `Lax` on both trackers
* changes the `cookieName` for `biz1` tracker

### DIFF
--- a/snowplow.js
+++ b/snowplow.js
@@ -15,17 +15,19 @@ import Cookies from 'js-cookie'
 import { COOKIE_PREF_KEY, DOCS_SITE_URLS } from './src/constants/config'
 import { reloadOnce } from './src/helpers/reloadOnce'
 
-const setupBrowserTracker = () => {
+const createTrackerConfig = (cookieName) => {
   const appId = DOCS_SITE_URLS.includes(window.location.hostname)
     ? 'docs2'
     : 'test'
   const domain = location.host.split('.').reverse()
 
-  const trackerConfig = {
+  let trackerConfig = {
     appId,
+    eventMethod: 'post',
     plugins: [LinkClickTrackingPlugin()],
     cookieDomain: `.${domain[1]}.${domain[0]}`,
-    cookieName: '_sp5_',
+    cookieName,
+    cookieSameSite: 'Lax',
     contexts: {
       webPage: true,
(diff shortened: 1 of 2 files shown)


igneel64 (approved):
LGTM

snowplow.js
```diff
@@ -15,17 +15,19 @@ import Cookies from 'js-cookie'
...
 
-const setupBrowserTracker = () => {
+const createTrackerConfig = (cookieName) => {
   const appId = DOCS_SITE_URLS.includes(window.location.hostname)
     ? 'docs2'
     : 'test'
   const domain = location.host.split('.').reverse()
 
-  const trackerConfig = {
+  let trackerConfig = {
```
stanch, line 24: why?

snowplow.js
```diff
@@ -41,8 +43,12 @@ const setupBrowserTracker = () => {
...
   }
 
-  newTracker('snplow5', 'https://collector.snowplow.io', trackerConfig)
-  newTracker('biz1', 'https://c.snowplow.io', trackerConfig)
+  return trackerConfig;
+};
+
+const setupBrowserTracker = () => {
+  newTracker('snplow5', 'https://collector.snowplow.io', createTrackerConfig('_sp5_'))
+  newTracker('biz1', 'https://c.snowplow.io', createTrackerConfig('_sp_biz1_'))
```
stanch, line 51: I think we should add the new cookie here: https://github.com/snowplow/documentation/blob/main/src/pages/cookie-preferences.mdx?plain=1#L64

stanch (changes requested):

snowplow.js
matus-tomlein: I misunderstood what the const keyword means. I thought it meant that the object is a constant and can't be changed (which it is on line 41 below). But reading more about it I see that it only prevents re-assigning and redeclaration. Good to know!

snowplow.js
matus-tomlein: Good point, done!

stanch (approved):

PR #166, opened 2022-10-31
Update manual event recovery doc

stanch:


stanch:
The example [here](https://github.com/snowplow-incubator/snowplow-event-recovery/blob/master/scripts/test.sc#L13) is broken (fails on line 13). If we could update that, then I would prefer to link there instead of keeping a copy of the script in the docs (which might become outdated).

For now I took the liberty to adapt the example provided by @ihortom. Also simplified things a little, as I don’t think the target users know enough Scala to make sense of the extra bits of information we were giving them.

peel (approved):

PR #172, opened 2022-11-03
Update pages for Python 0.12.0

Jack-Keene:


docs/collecting-data/collecting-from-own-applications/python-tracker/emitters/index.md
```diff
@@ -22,16 +22,19 @@ This is the signature of the constructor for the base Emitter class:
...
-        method: Literal["get", "post"] = "get",
+        method: Literal["get", "post"] = "post",
         buffer_size: Optional[int] = None,
         on_success: Optional[Callable[[PayloadDictList], None]] = None,
         on_failure: Optional[Callable[[int, PayloadDictList], None]] = None,
         byte_limit: Optional[int] = None,
         request_timeout: Optional[Union[float, Tuple[float, float]]] = None) -> None:
 ```
 
+**_\*\*Prior to v0.12.0_**
```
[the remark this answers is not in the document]
stanch: Make it an `:::note` maybe?

docs/collecting-data/collecting-from-own-applications/python-tracker/adding-extra-data-the-subject-class/index.md
```diff
@@ -191,7 +202,7 @@ You may want to track more than one subject concurrently. To avoid data about on
 ```python
 from snowplow_tracker import Subject, Emitter, Tracker
 
-# Create a simple Emitter which will log events to http://d3rkrsqld9gmqf.cloudfront.net/i
+# Create a simple Emitter which will log events to https://d3rkrsqld9gmqf.cloudfront.net/com.snowplowanalytics.snowplow/tp2
```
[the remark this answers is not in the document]
Jack-Keene: This is just the default URL that events will get sent to with the new default emitter settings, it's not customisable without changing the method/protocol settings so I don't think we need to explain this more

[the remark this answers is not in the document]
Jack-Keene: Oh wait, do you mean update `e = Emitter("d3rkrsqld9gmqf.cloudfront.net")` to something like `e = Emitter("{Collector URL}")`?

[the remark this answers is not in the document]
stanch: yes

docs/collecting-data/collecting-from-own-applications/python-tracker/emitters/index.md
```diff
@@ -22,16 +22,21 @@ This is the signature of the constructor for the base Emitter class:
...
         on_failure: Optional[Callable[[int, PayloadDictList], None]] = None,
         byte_limit: Optional[int] = None,
         request_timeout: Optional[Union[float, Tuple[float, float]]] = None) -> None:
 ```
 
+:::note
+Prior to v0.12.0
+:::
+Before version 0.12.0 the default values for the emitter `protocol` and `method` were `http` and `get` respectively.
+
```
stanch, line 39:
```suggestion
:::note Prior to v0.12.0

Before version 0.12.0 the default values for the emitter `protocol` and `method` were `http` and `get` respectively.

:::
```

docs/collecting-data/collecting-from-own-applications/python-tracker/emitters/index.md
```diff
@@ -46,7 +51,7 @@ def __init__(
 
 - `protocol`
 
-`protocol` defaults to "http" but also supports "https".
+`protocol` defaults to "https" but also supports "http". **_\*\*(Since v0.12.0_)**
```
stanch, line 54:
```suggestion
`protocol` defaults to "https" but also supports "http".

:::note Prior to v0.12.0

In older versions, `protocol` defaulted to `http`.

:::
```

matus-tomlein: Feels like a bit too much having this note right after the previous note which said the same thing.

docs/collecting-data/collecting-from-own-applications/python-tracker/emitters/index.md
```diff
@@ -22,16 +22,21 @@ This is the signature of the constructor for the base Emitter class:
...
 ```
 
+:::note Prior to v0.12.0
+
+Before version 0.12.0 the default values for the emitter `protocol` and `method` were `http` and `get` respectively.
+
+:::
 | **Argument** | **Description** | **Required?** | **Type** |
 | --- | --- | --- | --- |
 | `endpoint` | The collector URI | Yes | Dict |
```
matus-tomlein, line 42: Nit: could we add a column with the default values similar like in the [tracker parameters](https://docs.snowplow.io/docs/collecting-data/collecting-from-own-applications/python-tracker/intialization/#creating-a-tracker)?

matus-tomlein: Also can you please fix the next row with protocol – I don't think one can use capital `HTTP` and `HTTPS` as it suggests. Should be something like: `Request protocol: "http" or "https"`. Also the type should not be a `List` but `String`

matus-tomlein: And the type for the `endpoint` should be a `String` as well.

matus-tomlein (approved):
LGTM!

docs/collecting-data/collecting-from-own-applications/python-tracker/adding-extra-data-the-subject-class/index.md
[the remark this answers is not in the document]
matus-tomlein: This is a good idea but it hides the expected format of the addres, could we make it something like this?

```
e = Emitter("collector.example.com")
```

PR #175, opened 2022-11-06
Delete reference to "beam enrich" as it's deprecated

daisy1754:
https://discourse.snowplow.io/t/rfc-beam-enrich-deprecation/5602


stanch (approved):
Many thanks!

PR #194, opened 2022-11-21
Add BQ Loader quickstart changes

jbeemster:
@stanch the only thing I think is missing is:

1. The diagram on the BigQuery one -> I don't have the source diagram parts so that needs to be updated to reflect new flow
2. Not something we can do now but we have no bad_row flow for BigQuery loading due to the fact that we only have a GCS Loader that runs on top of Beam.  We need something more like the S3 Loader but Beam apps are not something I want to support in quickstart.


docs/getting-started-on-snowplow-open-source/quick-start-gcp/query-your-data/index.md
```diff
@@ -0,0 +1,139 @@
...
+
+:::caution
+
+Currently the Quick Start guide pipelines do not load bad data into BigQuery directly.
+
+:::
+
+To query it, you will need to follow a few additional steps manually:
+
+- Configure the [GCS Loader](/docs/pipeline-components-and-applications/loaders-storage-targets/google-cloud-storage-loader/index.md) to load bad data from the `Bad Rows` topic into GCS.
```
stanch, line 124: @jbeemster or @istreeter could you confirm these steps?

jbeemster: So not just bad_rows but also the bad stream.

You have one bad stream coming from enrich and another coming from the bigquery loader.  So you would need a GCS Loader for both.  Generally when we talk about bad data we are talking about enrich.

stanch (approved):

PR #201, opened 2022-11-24
Unify the 'what is deployed' pages for AWS

stanch:
Now it's a single tabbed page. Also updated the images so that they don’t jump when switching tabs.


mharvey83 (approved):
LGTM

PR #204, opened 2022-11-29
Move to more nested structure, remove tabs where possible

rlh1994:
Simplify the structure of the dbt notes, removing lots of tabs where it's not strictly necessary, and prepare us for a growth in the number of dbt packages.


agnessnowplow (approved):
LGTM!

PR #212, opened 2022-12-05
Add the terms “canonical event model” to the relevant page

stanch:
This helps users find the page when they search for “canonical event model”


adatzer (approved):
Looks good to me!

PR #213, opened 2022-12-07
Update BQ loader version to 1.6.1

istreeter:


mharvey83 (approved):
Happy with the change.

PR #215, opened 2022-12-08
Tidy dbt configuration list

rlh1994:
- Add more of the variables (most of them get defined in the quick start but weren't in this table)
- Sorts the table(s) alphabetically to make it easier to find the config you are looking for
- Remove the `snowplow__` prefix in the table to make it less wide
- Fixes some hard links in the quickstart to be relative


emielver (approved):
LGTM!

PR #217, opened 2022-12-09
Add IDFA documentation for mobile trackers

matus-tomlein:
This PR adds a missing section to the mobile trackers documentation that explains how to configure tracking of the advertising identifier (IDFA).

I also added instructions for the v5 of iOS tracker which is not yet out but the second alpha release will be out in the coming weeks so plan to merge it in then.


mscwilson (commented):
The info about IDFA etc looks great. I think it would work well as a separate page. That would give us more space for explanation if we add other identifiers.

I think the Snowplow/mobile docs in general would be easier to use if there were more, shorter, specific pages.

mscwilson (approved):
Great info, big improvement for the mobile docs.

PR #223, opened 2022-12-15
Add postgres destination in the manual GCP setup

lmath:
We specify BigQuery Loader and GCS loader as potential storage targets for a manual GCP pipeline setup, but do not mention the Postgres loader, so we add it here.

I set up the postgres loader for a GCP pipeline following the existing instructions, which seem to be up to date, but I thought it would be helpful to specify that although the loader will create the events table, that the database and schema supplied in the config do need to be created to be able to start the loader.

### DIFF
--- a/docs/getting-started-on-snowplow-open-source/setup-snowplow-on-gcp/setup-postgres-destination/index.md
+++ b/docs/getting-started-on-snowplow-open-source/setup-snowplow-on-gcp/setup-postgres-destination/index.md
@@ -1,0 +1,8 @@
+---
+title: "Setup Postgres Destination"
+sidebar_position: 70
+---
+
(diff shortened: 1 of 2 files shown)


stanch:
Two things I’d like to change here:
1. Postgres is also supported and missing on AWS, I think we should add it there as well?
2. On AWS, we group the destinations into their own category in the left pane, which looks nice. On GCP we don’t but should. Introducing another GCP destination page before we do the grouping means that later when we change this we’ll need to maintain an extra redirect. I’d rather first move the existing pages and then create the new page. (Could be in the same PR though.)

lmath:
@stanch - I'm going to close this PR in favour of https://github.com/snowplow/documentation/pull/226 which will include these changes.

PR #230, opened 2022-12-19
Update Rust tracker docs for 0.2.0

greg-el:
The PR adds docs for the release of [0.2.0 ](https://github.com/snowplow/snowplow-rust-tracker/pull/34)of the Rust tracker.


docs/collecting-data/collecting-from-own-applications/rust-tracker/getting-started/index.md
```diff
@@ -31,7 +31,7 @@ use snowplow_tracker::Snowplow;
 Instantiate a tracker using the `Snowplow::create_tracker` function. The function takes three required arguments: `namespace`, `app_id`, `collector_url`, and one optional argument, `subject`. Tracker `namespace` identifies the tracker instance; you may create multiple trackers with different namespaces. The `app_id` identifies your app. The `collector_url` is the URI of the Snowplow collector to send the events to. `subject` allows for an optional Subject to be attached to the tracker, which will be sent with all events.
 
 ```rust
-let tracker = Snowplow::create_tracker("ns", "app_id", "https://...", None);
+let mut tracker = Snowplow::create_tracker("ns", "app_id", "https://...", None);
 ```
 
 There are additional optional arguments to configure the tracker. To learn more about configuring how events are sent, check out [this page](/docs/collecting-data/collecting-from-own-applications/rust-tracker/initialization-and-configuration/index.md).
```
matus-tomlein, line 37: Nit: would it make sense to also give a snippet how to deinitialize the tracker using the `close_emitter()` call here? I know it's mentioned on the following page but it's kind of buried and might be missed by some users.

matus-tomlein (approved):
LGTM!

I don't know how I feel about duplicating the 0.1.0 docs under `previous-versions`. It seems a bit of an overkill since there are not that many breaking changes and we are still in the early releases. Maybe a short migration page would be enough to show what changed? It's quite annoying that when I search for something in the docs, it shows the results multiple times for each copy in previous versions...

stanch (changes requested):
You need to add a root `index.md` for `previous-versions` so that it displays as `Previous versions` — see other trackers for how it’s done. Also I have a slight preference for it to appear _before_ the API docs link.

image

greg-el:
I've removed the `previous-versions` anyway as Matus suggested, as the breaking changes are minimal. I've added a short migration guide "caution" on the index page for the tracker.

docs/collecting-data/collecting-from-own-applications/rust-tracker/getting-started/index.md
greg-el: Yeah, this seems like a good idea.

docs/collecting-data/collecting-from-own-applications/rust-tracker/index.md
```diff
@@ -14,6 +14,14 @@ The Snowplow Rust Tracker allows you to add analytics to your Rust apps.
 
 The tracker is published on crates.io as [snowplow_tracker](https://crates.io/crates/snowplow_tracker).
 
+:::caution Migration from 0.1.0
+
+If you are upgrading from 0.1.0 to 0.2.0, there are a couple of changes required to make to your code:
+
+- [`Tracker.track`](./getting-started/#tracking-events) is no longer an async function
```
stanch, line 21:
```suggestion
- [`Tracker.track`](./getting-started/index.md#tracking-events) is no longer an async function
```

stanch (approved):

PR #236, opened 2022-12-30
Use Postgres and BigQuery terraform.tfvars files

jbsilva:
Both files have to be used in order to read the `postgres_db_enabled` and `bigquery_db_enabled` variables.

### DIFF
--- a/docs/getting-started-with-snowplow-bdp/enterprise/setup-guide-aws/index.md
+++ b/docs/getting-started-with-snowplow-bdp/enterprise/setup-guide-aws/index.md
@@ -53,7 +53,6 @@ You will need to share this role with us as part of filling out the setup form i
 
 ```text
  "acm:*",
- "apigateway:*",
  "application-autoscaling:*",
  "autoscaling:*",
  "aws-marketplace:Subscribe",
@@ -66,7 +65,6 @@ You will need to share this role with us as part of filling out the setup form i
  "ec2:*",
  "ecs:*",
  "eks:*",
- "elasticbeanstalk:*",
  "elasticloadbalancing:*",
  "elasticmapreduce:*",
  "es:*",


snowplowcla:
Thanks for your pull request. Is this your first contribution to a Snowplow open source project? Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).

:memo: Please visit https://docs.snowplowanalytics.com/docs/contributing/contributor-license-agreement/ to learn more and sign.

Once you've signed, please reply here (e.g. I signed it!) and we'll verify. Thanks.

jbsilva:
> Thanks for your pull request. Is this your first contribution to a Snowplow open source project? Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> 📝 Please visit https://docs.snowplowanalytics.com/docs/contributing/contributor-license-agreement/ to learn more and sign.
> 
> Once you've signed, please reply here (e.g. I signed it!) and we'll verify. Thanks.

I signed it!

snowplowcla:
Confirmed! @jbsilva has signed the Contributor License Agreement. Thanks so much.

jbeemster:
Thanks @jbsilva for the contribution.  I assume the want to use both vars files comes from when using `bigquery` only you then get prompted for `var.postgres_db_name`, `var.postgres_db_password` and `var.postgres_db_username`?

If that's the case the better solution might instead be to set a default value in the quickstart directly so you could either pick bigquery OR postgres and not worry about competing variable values between these two values.

Or is your use-case here that you want to enable both loaders?

jbsilva:
This pull request is just to update the documentation.

**The outputs changed** and now there is no `terraform.tfvars` file, but a `postgres.terraform.tfvars` and a `bigquery.terraform.tfvars`.

I added the `-var-file` because I believe that a simple `terraform apply` won't read the files with their current name.

From Terraform's documentation (https://developer.hashicorp.com/terraform/language/values/variables#variable-definition-precedence):

> Terraform loads variables in the following order, with later sources taking precedence over earlier ones:
> 
>     Environment variables
>     The terraform.tfvars file, if present.
>     The terraform.tfvars.json file, if present.
>     Any *.auto.tfvars or *.auto.tfvars.json files, processed in lexical order of their filenames.
>     Any -var and -var-file options on the command line, in the order they are provided.

stanch:
Hi @jbsilva, in the meantime we’ve clarified this in a slightly different way — does https://github.com/snowplow/documentation/pull/269 work for you?

jbsilva:
Hi @stanch, much better now!
Thanks for making everything clearer.

BTW, is it possible to enable both PostgreSQL and BigQuery?

stanch:
Don’t see why not, but will let @jbeemster confirm. Out of interest, what’s your use case for running both? :)

jbsilva:
> Don’t see why not, but will let @jbeemster confirm. Out of interest, what’s your use case for running both? :)

The main reason for now is that I'm setting up an environment for training and it would be nice to show both formats.

I see there are some differences between them and some tools (the Snowplow E-Commerce dbt package, for example) only work with one of them.

PR #249, opened 2023-01-09
Rework custom model page

rlh1994:
The current custom model page can be somewhat confusing in exactly how to build a custom model and what is required. This re-works the page to make clear the 3 types of custom model people could build, and how to work with each one in terms of creating, backfilling, and restarting.

Also fixes a broken link and adds jinja syntax support


agnessnowplow (approved):
LGTM!

stanch (approved):
I’m out of my depth on the actual content, but this looks like a good change for readability. The whole module vs model language had me confused in the past.

PR #256, opened 2023-01-11
Remove adapter restriction notification box

agnessnowplow:
As the name suggests, quick update to remove restriction notification to Snowflake/BigQuery as we now support all (the same) adapters as the main web model.


docs/modeling-your-data/modeling-your-data-with-dbt/dbt-models/dbt-web-data-model/index.md
```diff
@@ -27,11 +27,6 @@ This model consists of a series of modules, each producing a table which serves
 ## Optional Modules
 
 ### Consent Tracking Custom Module
-
```
stanch, line 30: This seems to leave an empty gap between two headers of the same level. Should the Introduction be a `####` instead?

agnessnowplow: True, and tbh, the Introduction header seems superfluous as there wasn't any for the main web mode above either, I will remove it altogether.

rlh1994 (approved):

emielver (approved):

PR #271, opened 2023-01-20
Make clear resolver url

rlh1994:


rlh1994:
@istreeter I like it, wasn't sure what that part is called but that helps make things a lot clearer. I've also made a few other tweaks such as bumping the version. I'm not 100% on exactly how the ttl works so that might not be entirely correct.

PR #275, opened 2023-01-23
Python 0.13.0

Jack-Keene:


docs/collecting-data/collecting-from-own-applications/python-tracker/emitters/index.md
```diff
@@ -22,108 +22,90 @@ This is the signature of the constructor for the base Emitter class:
...
 
-Since version 0.10.0, the constructors takes another `request_timeout` argument.
+The default buffer capacity is 10,000 events. This is the number of events that can be stored. When the buffer is full, new tracked payloads are dropped, so choosing the right capacity is important. You can set the buffer capacity through the `EmitterConfiguration` object, for example:
 
-:::
-- `request_timeout`
-
-Timeout for HTTP requests. Can be set either as single float value which applies to both "connect" AND "read" timeout, or as tuple with two float values which specify the "connect" and "read" timeouts separately.
+```python
+
```
stanch, line 99: Unnecessary indentation

stanch: (I think I also saw somewhere else, could you check?)

stanch (commented):
Formatting-wise, looks fine aside from indentation. But aside from that I don’t feel qualified to review :)

docs/collecting-data/collecting-from-own-applications/python-tracker/emitters/index.md
```diff
@@ -22,108 +22,90 @@ This is the signature of the constructor for the base Emitter class:
 def __init__(
         self,
         endpoint: str,
-        protocol: Literal["http", "https"] = "https",
+        protocol: HttpProtocol = "https",
```
matus-tomlein, line 25: I think this was more useful in the previous form (`Literal["http", "https"]`) because now the reader doesn't know what the `HttpProtocol` type alias represents. The same for `Method`, `SuccessCallback`, and `FailureCallback`.

docs/collecting-data/collecting-from-own-applications/python-tracker/tracking-specific-events/index.md
```diff
@@ -262,6 +267,30 @@ Example:
 ```python
 t.track_screen_view("HUD > Save Game", "screen23", None, 1368725287000)
 ```
+
+### Track mobile screen view
```
matus-tomlein, line 271: Could you mention the schema here?

docs/collecting-data/collecting-from-own-applications/python-tracker/tracking-specific-events/index.md
```diff
@@ -245,6 +246,10 @@ t.track_page_ping("http://mytesturl/test2", "Page title 2", "http://myreferrer.c
 
 ### Track screen view
 
+:::note
```
matus-tomlein, line 249: Also would be nice to link the schema here (maybe a link to Github) and also add `(deprecated)` to the section title to make it more clear.

matus-tomlein: Actually, maybe it would be better to just mention this deprecated screen view API as a note under the `Track mobile screen view` section (that could also be renamed to just `Track screen view`)? A bit worried about confusing people with having two similar sections and we want to direct people to the new API.

docs/collecting-data/collecting-from-own-applications/python-tracker/emitters/index.md
```diff
@@ -22,108 +22,90 @@ This is the signature of the constructor for the base Emitter class:
...
+See the [`API docs`](https://snowplow.github.io/snowplow-python-tracker/) for more information on the individual parameters. 
+## What happens if an event fails to send?
+:::note New in v0.13.0
+Retry capabilities are new in v0.13.0
 :::
+After trying to send a batch of events the collector will return an http status code. A 2xx code is always considered successful. If a failure code is returned (anything other than 2xx, with certain exceptions, see below), the events (as PayloadDictList objects) are returned to the buffer. They will be retried in future sending attempts. 
 
-- `buffer_size`
-
-When the emitter receives an event, it adds it to a buffer. When the queue is full, all events in the queue get sent to the collector. The `buffer_size` argument allows you to customize the queue size. By default, it is 1 for GET requests and 10 for POST requests. (So in the case of GET requests, each event is fired as soon as the emitter receives it.) If the emitter is configured to send POST requests, then instead of sending one for every event in the buffer, it will send a single request containing all those events in JSON format.
```
matus-tomlein, line 64: Was it intentional that you removed these docs about the `batch_size`, `byte_limit`, `on_success`, `request_timeout` and other properties? I thought they were quite well written, wondering if we can keep some of it.

matus-tomlein (changes requested):
Could we also add a migration guide with the list of the breaking changes and how should users update their apps (e.g., the `batch_size`, retry behavior – users should no longer retry in the `on_failure` callback as the docs suggested before)? Maybe something [like this page on the C++ tracker](https://docs.snowplow.io/docs/collecting-data/collecting-from-own-applications/c-tracker/upgrading-to-newer-versions/)

docs/collecting-data/collecting-from-own-applications/python-tracker/emitters/index.md
Jack-Keene: I removed these just because it seemed quite similar to the API docs. I'll add them back in, but I think it could be good to re-structure this page for V1 though?

matus-tomlein: Feel free to move it around, I am just worried about losing the information. The API docs can be quite terse, but I think we can be more verbose in these docs and explain things in a more understandable way to users which I think that text did well.

PR #276, opened 2023-01-23
Move Collector & Enrich to fitting sections

stanch:
Recently we moved the Loaders to the “Storing (...) data” section. Much in the same spirit, this commit moves Collector and Enrich to move fitting places. This helps the reader understand the context around the apps.

I ran the following:
```
./move.sh docs/pipeline-components-and-applications/stream-collector docs/collecting-data/stream-collector
./move.sh docs/pipeline-components-and-applications/enrichment-components docs/enriching-your-data/enrichment-components
```
and then adjusted the titles and positions by hand.


benjben (commented):
`Enriching your data` => We might want to make it more clear that Enrich is not only about enriching, but also validating

paulboocock:
One of the reasons we had the pipeline components slightly more (ok, a lot more) hidden away, we to reduce potential confusion for BDP users. Historically, we used to get quite a few support tickets from people who end up confused about what they need to do as they read through the various docs pages and some even ended up trying to spin up their own collectors (I'll never quite be sure why).

Now, don't get me wrong, I think there's a place for collector and enrichment documentation within these sections. However, I don't think it's these specific pages (which are more setup/reference guides). What I feel we need here, is a whats what, how does this tech work sort of thing, rather than a heres a command to run to pull it from docker hub and here's a load of config settings that you can't access directly.

If we do want to bring this level of complexity into this section, then it almost certainly needs splitting up into a Reference guide area and a configuration guide area. Pages [like this one](https://deploy-preview-276--snowplow-docs.netlify.app/docs/using-the-snowplow-console/accessing-collector-configuration/) should probably be kept with the collecting data bit if thats where we're going (but it also makes sense to group all the Console UI pages together too 😬)

stanch:
Valid points, let me rethink!

PR #282, opened 2023-01-25
Move downloading and running section from README into Igluctl docs

lmath:
There was a fair bit of duplication in Igluctl's README and here in the Igluctl docs. We shouldn't have duplicated info either way, but I think it makes sense for info about running/using Igluctl can live in docs, and info about developing Igluctl can live in the README.

I raised a [PR in Igluctl](https://github.com/snowplow/igluctl/pull/138) to remove duplication in the README. It wasn't all duplication. There was a point about how running Igluctl is different if you're on Windows vs OS X/Linux, so I've moved that point over to the docs. 

Here is the section that has been moved over:
image

### DIFF
--- a/docs/pipeline-components-and-applications/iglu/igluctl-2/index.md
+++ b/docs/pipeline-components-and-applications/iglu/igluctl-2/index.md
@@ -4,6 +4,11 @@ date: "2021-01-21"
 sidebar_position: 10
 ---
 
+```mdx-code-block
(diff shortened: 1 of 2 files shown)


benjben (approved):
:+1: Thanks @lmath !

I'll ping you when I release `0.10.2`

benjben:
It's been released @lmath

PR #285, opened 2023-01-26
Bump media player version

emielver:
Bumping


rlh1994 (approved):
As the ghostbusters nearly said, bumping makes me feel good 🎷

PR #288, opened 2023-01-27
BigQuery loader: remove dropwizard from configuration reference

benjben:
These parameters are deprecated since we remove Dataflow job from the codebase.


lmath (approved):
Good stuff 👍

PR #298, opened 2023-02-01
Set Enrich version to 3.7.0

benjben:


stanch (changes requested):
Could you add a small note here somewhere https://docs.snowplow.io/docs/pipeline-components-and-applications/enrichment-components/configuration-reference/ that environment variables are supported since 3.7.0?

benjben:
@stanch environment variables were already supported in the hocon, so I put the info in another place

docs/pipeline-components-and-applications/enrichment-components/enrich-kafka/index.md
```diff
@@ -31,6 +31,8 @@ Above assumes that you have following directory structure:
 2. Iglu Resolver [configuration JSON](/docs/pipeline-components-and-applications/iglu/iglu-resolver/index.md)
 3. [configuration HOCON](/docs/pipeline-components-and-applications/enrichment-components/configuration-reference/index.md)
 
+It is possible to use environment variables in all of the above.
```
stanch, line 34: But for enrichments and Iglu this is only available since 3.7.0 right? I would clarify that.

benjben: Good point ! I updated

stanch (approved):

PR #303, opened 2023-02-08
Update tab container css

rlh1994:
This makes 2 changes:
1) reduced the vertical padding on tab group headers to 10px either side instead of 16 as, at least to me, this felt a bit bloated especially if the tab'd content was only a few lines
2) Adds a left border and minor left padding to the entire tab container so it is easier to identify where the tab content starts and stops (I will accept changes where the border is offset and the container doens't have the left padding, but to be honest I couldn't work out how to do that so I am going to say this looks better).

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -64,6 +64,7 @@
   --ifm-toc-padding-horizontal: 20px;
   --docusaurus-highlighted-code-line-bg: rgba(0, 0, 0, 0.1);
   --doc-sidebar-width: 330px !important;
+  --ifm-tabs-padding-vertical: 7px;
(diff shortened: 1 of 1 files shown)


src/css/custom.css
```diff
@@ -303,3 +304,33 @@ textarea {
...
+
+.tabs-container  {
+  padding-left: 14px;
+  padding-right: 14px;
+  padding-bottom: 14px;
+  padding-top: 7px;
+  border: 2px outset transparent;
+  border-radius: var(--ifm-alert-border-radius);
+  transition: 1.5s ease-in-out all;
+}
```
stanch, line 336:
```suggestion
.tabs-container:hover  {
  border: 2px solid var(--ifm-color-primary);
}

.tabs-container  {
  padding: 7px 14px 14px;
  border: 2px solid transparent;
  border-radius: var(--ifm-alert-border-radius);
  transition: 0.5s ease-out all;
}
```

stanch (approved):

stanch (approved):

PR #304, opened 2023-02-08
Feature/add fractribution updates

agnessnowplow:
Adding some updates for the upcoming 0.2.0 release of the `snowplow_fractribution` data model.

### DIFF
--- a/docs/modeling-your-data/modeling-your-data-with-dbt/index.md
+++ b/docs/modeling-your-data/modeling-your-data-with-dbt/index.md
@@ -83,7 +83,8 @@ The supported data warehouses per version can be seen below:
 
 | snowplow-fractribution version | dbt versions      | BigQuery | Databricks | Redshift | Snowflake | Postgres |
 | ----------------------------- | ----------------- | -------- | ---------- | -------- | --------- | -------- |
-| 0.1.0                         |  >=1.0.0 to <2.0.0 | ✅       |  ✅       | ❌       | ✅       | ❌        |
+| 0.2.0                         |  >=1.3.0 to <2.0.0 | ✅       |  ✅       | ❌       | ✅       | ❌        |
+| 0.1.0                         |  >=1.0.0 to <2.0.0 | ❌      |  ❌       | ❌       | ✅       | ❌        |
 
 </TabItem>
 </Tabs>
--- a/docs/modeling-your-data/modeling-your-data-with-dbt/index.md
+++ b/docs/modeling-your-data/modeling-your-data-with-dbt/index.md
@@ -83,7 +83,7 @@ The supported data warehouses per version can be seen below:
 
 | snowplow-fractribution version | dbt versions      | BigQuery | Databricks | Redshift | Snowflake | Postgres |
 | ----------------------------- | ----------------- | -------- | ---------- | -------- | --------- | -------- |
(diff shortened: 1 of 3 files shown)


docs/modeling-your-data/modeling-your-data-with-dbt/index.md
```diff
@@ -83,7 +83,7 @@ The supported data warehouses per version can be seen below:
 
 | snowplow-fractribution version | dbt versions      | BigQuery | Databricks | Redshift | Snowflake | Postgres |
 | ----------------------------- | ----------------- | -------- | ---------- | -------- | --------- | -------- |
-| 0.1.0                         |  >=1.0.0 to <2.0.0 | ❌       |  ❌       | ❌       | ✅       | ❌        |
+| 0.1.0                         |  >=1.0.0 to <2.0.0 | ✅       |  ✅       | ❌       | ✅       | ❌        |
```
rlh1994, line 86: bump version number

rlh1994 (changes requested):
Looks good other than the version number needs bumping. I also can't remember what version of our utils 0.1.0 was on, but 0.2.0 will definitely only support dbt 1.3.0 or greater, double check if 0.1.0 could support lower?

zloff:
I'm a little confused about the difference in purpose between the `dbt-fractribtuion-data-model/index.md` file and the Fractribution section in the `dbt-quickstart/index.md` file.

What is the aim of each document? From what I read above, where we are linking to the quickstart for 'more detail' it seems like the quickstart is supposed to be more detailed, but when I look at an example like snowplow_web, the quickstart section is much less detailed than the modelling page. I can just imagine as a user being frustrated by clicking through to something purportedly with more detail, that doesn't actually give me that. Or if I was reading the docs for two different packages, and the info in the quickstart was the most detailed for one package, but then the documentation in the modelling section was the most detailed for the other package.

zloff:
I also just realised that we haven't mentioned the docker image anywhere except in the accelerator. Is that for any specific reason, or did we just forget? I can add information about it if we haven't left it out for any specific reason. I figure we should mention it as an option instead of running the python script, probably a lot of people would prefer it since you don't have to make any env changes

rlh1994:
> I'm a little confused about the difference in purpose between the dbt-fractribtuion-data-model/index.md file and the Fractribution section in the dbt-quickstart/index.md file.

The line is getting blurrier with the more packages we add and the more complex they are (compared to web and mobile for example which have a lot of overlap). We used to have each package have fully separate docs but there was so much overlap this seemed cleaner.

My personal opinion is quick start is the bare minimum I need to do to realistically get it running. Explanations about specific configurations or custom ways to do stuff would go in the specific package page (or the configuration page). 

> I also just realised that we haven't mentioned the docker image anywhere except in the accelerator. 

No reason I know of, makes sense to add it in.

agnessnowplow:
Now that we clarified the section goals I reduced the quick start guide in places where Ryan pointed it out, it is hard to draw the line between what is essential and what is a rather good-to-know, as the operation of this packages is not so straightforward so better be explicit or repeat ourselves than leave the user confused. @zloff Can you please add instructions on using the docker, please?

rlh1994:
I think I've copied everything here over to #309 (because I was going to have a painful rebase otherwise), @zloff can you check if there's anything wrong please?

PR #311, opened 2023-02-12
Update stream-collector to 2.8.2

jbsilva:
Use newest version in documentation


stanch:
Thanks @jbsilva!

Throughout the docs, we use component versions as variables, but these pages somehow got left out. May I ask you to use that approach? Example (from https://github.com/snowplow/documentation/blob/main/docs/pipeline-components-and-applications/stream-collector/setup/index.md?plain=1#L33)

````
```mdx-code-block
import {versions} from '@site/src/componentVersions';
import CodeBlock from '@theme/CodeBlock';
```

<CodeBlock language="bash">{
`docker run --rm \\
  -v $PWD/config.hocon:/snowplow/config.hocon \\
  -p 8080:8080 \\
  snowplow/scala-stream-collector-\${flavour}:${versions.collector} --config /snowplow/config.hocon
`}</CodeBlock>
````

On a different note, I wonder if we should be recommending to set up Collector as described here, manually downloading and running the jar. Would it be better to point the reader to https://docs.snowplow.io/docs/pipeline-components-and-applications/stream-collector/setup/?

stanch (changes requested):
Use `componentVersions.js`

stanch:
Will be solved in #349.

PR #326, opened 2023-03-01
Add basic feedback widget

mscwilson:
A basic user feedback widget. The feedback is sent to us using Structured events.

NB: the `DocItem/Footer` is "unsafe" to swizzle. Might break in minor (not just major) versions of Docusaurus.


colmsnowplow:
👋 Two quick Qs - 1. where should I expect to see the widget (on the deploy preview), 

2. In the structured events we're sending, what's the link? My assumption is that someone submits feedback, that's recorded somewhere, and the link is to a link to that? 

Sorry if they're basic questions, I'm having a slow day so might be missing obvious things!

greg-el:
Not sure if there's anything built-in to do this for you, but we probably want `cursor: pointer` on the buttons so they're more obviously clickable. 

@colmsnowplow if you scroll down to the bottom of a page, it should be on the left above the "edit this page" link

src/theme/DocItem/Footer/index.js
```diff
@@ -0,0 +1,196 @@
...
+    trackStructEvent({
+      category: 'feedback',
+      action: 'dislike',
+      label: permalink,
+    })
+  }
+
+  const handleSubmit = (event) => {
+    event.preventDefault()
+    console.log('submitted')
```
greg-el, line 94: 👀

mscwilson:
>     2. In the structured events we're sending, what's the link? My assumption is that someone submits feedback, that's recorded somewhere, and the link is to a link to that?

The `permalink` bit? It's the link for the docs page, so we know which one they were on.

colmsnowplow:
ty for explaining! Looks good!

rlh1994:
Is there a way to align the text and buttons to start in line with the rest of the document? The right side is aligned but the left starts further in? 
image

mscwilson:
> Is there a way to align the text and buttons to start in line with the rest of the document? The right side is aligned but the left starts further in?

CSS 😤 I already aligned that once! Will try again

src/theme/DocItem/Footer/index.js
```diff
@@ -0,0 +1,206 @@
...
+              formattedLastUpdatedAt={formattedLastUpdatedAt}
+              lastUpdatedBy={lastUpdatedBy}
+            />
+          )}
+        </div>
+      </div>
+    </div>
+  )
+}
+export default function DocItemFooter() {
```
stanch, line 176: Would it be possible to wrap instead of ejecting? (`npm run swizzle @docusaurus/theme-classic DocItem/Footer -- --wrap --danger`) That way we can minimize the potential for breaking changes. The feedback widget could go right above or right below the footer.

src/theme/DocItem/Footer/styles.module.css
```diff
@@ -0,0 +1,93 @@
+.lastUpdated {
```
stanch, line 1: Could you pick 1 naming style for all CSS classes? Either kebab-case or camelCase. Currently this file uses both camelCase and snake_case.

src/theme/DocItem/Footer/styles.module.css
```diff
@@ -0,0 +1,93 @@
...
+}
+
+.feedback_question {
+  font-size: 14px;
+  font-style: italic;
+  padding-right: 14px;
+  line-height: 40px;
+}
+
+.thanks {
```
stanch, line 20: I would use a more specific name :)

src/theme/DocItem/Footer/styles.module.css
```diff
@@ -0,0 +1,93 @@
...
+  margin-left: 0px;
+  font-size: 14px;
```
stanch, line 22: There are a few font sizes used in this file, and all are inline. Could we use a variable?

Also, I would try to make this more uniform:
image

I think the feedback stuff could use the same font size as the “Edit this page” link.

src/theme/DocItem/Footer/styles.module.css
```diff
@@ -0,0 +1,93 @@
...
+  margin-left: 16px;
+  box-sizing: border-box;
+  max-width: 80%;
+  border: 2px solid var(--ifm-color-primary);
+  border-radius: var(--ifm-alert-border-radius);
+  background-color: var(--ifm-navbar-background-color);
+  resize: none;
+}
+
+.feedback_button {
```
stanch, line 40: Can we not reuse `snwpl-nav-button`? And only override a few things here. Otherwise this includes lots of inline colors that we will forget to change if we change the top buttons.

src/theme/DocItem/Footer/styles.module.css
```diff
@@ -0,0 +1,93 @@
...
+  color: var(--ifm-font-color-base);
+}
+
+.feedback_button:active,
+.feedback_button:hover,
+.feedback_button:focus {
+  text-decoration: none;
+  background: #e4dcef;
+  border-color: #e4dcef;
+}
```
stanch, line 93: We should not need these if we reuse existing button classes.

stanch (changes requested):
There are a few wonky layouts that appear at different screen sizes:

image

image

My suggestion is:
* Simplify the code: wrap the footer and just put the feedback above or below, instead of squeezing it on the side.
* Simplify the layout: always put the comment box below the buttons. That will reduce the amount of CSS, margins, etc.

src/theme/DocItem/Footer/styles.module.css
stanch: And the horizontal/vertical margins should be the same.

src/theme/DocItem/Footer/index.js
```diff
@@ -0,0 +1,147 @@
...
+    setTimeout(() => {
+      setIsThanksVisible(false)
+    }, 1000)
+  }
+
+  return (
+    <div className={clsx('col', styles.feedback)}>
+      <div className={clsx('row', styles.feedback)}>
+        <div className={clsx('row', styles.feedbackQuestion)}>
+          Was this page helpful?
```
stanch, line 98: See if you have the spare minute to add this
image
as discussed on Slack

stanch (approved):

PR #328, opened 2023-03-02
Automatically explain abbreviations

stanch:


mscwilson (approved):
That's really cool. 

Nit: marking the abbreviations in headers seems like overkill, would it be easy to prevent it happening there?

stanch:
> Nit: marking the abbreviations in headers seems like overkill, would it be easy to prevent it happening there?

It _is_ easy:
```diff
-    flatMap(ast, node => {
-      if (node.type === 'text') {
+    flatMap(ast, (node, index, parent) => {
+      if (node.type === 'text' && parent.type !== 'heading') {
```

But I think I prefer not skipping them, because on some pages the header might be the only place where the abbreviation appears.

PR #331, opened 2023-03-07
Add placeholder pages for LaunchDarkly GTM-SS Tag

adatzer:
Not sure if best practice, the reason for doing so is to ensure the pages are there before the repository goes public.


adatzer:
Mainly to be able to check that the links (referenced in Readme, metadata) have no typo and we surely avoid having public links that don't exist.  Alternatively we could publish the final docs before making the repository public, but this is still work in progress and i am not sure of the exact order of events yet. Not a major issue, i just thought it may make sense to detach the links' existence from the rest of the things to be done. What do you think?

stanch (approved):
Sounds good!

PR #337, opened 2023-03-10
Add dbt and sql runner latest versions

rlh1994:
Adds versions of the dbt packages and the sql runner versions. 

So far as I can find we only hardcoded them in one place, the package overview table which required the work of a "genius" to solve (not my words).


src/componentVersions.js
```diff
@@ -1,3 +1,11 @@
+// To use this in other pages, add the following to the top of your page
+// ```mdx-code-block
+// import {versions} from '@site/src/componentVersions';
+// import CodeBlock from '@theme/CodeBlock';
+// ```
+// Then use <>{versions.yourThing}</> to add the version in place.
+// Note that this does not render correctly in markdown ables so you need to use additional steps, see docs/modeling-your-data/modeling-your-data-with-dbt/index.md as an example
```
stanch, line 7: There is some info about this in the README also. Can we improve the README (I think it has a missing link) and redirect the reader there instead?

stanch (changes requested):

stanch (approved):
Genius

PR #342, opened 2023-03-13
Enrich: remove KPL

benjben:
We dropped KPL but we still mention KPL metrics in the docs, as stated on [discourse](https://discourse.snowplow.io/t/enricher-kpl-metrics-in-cloudwatch/8262/2).


stanch (approved):
Could you add the meaning of KCL to `src/remark/abbreviations.js`?

stanch (approved):

PR #343, opened 2023-03-13
GCS loader: remove template method

benjben:
Template doesn't work any more (cf [discourse](https://github.com/snowplow-incubator/snowplow-google-cloud-storage-loader/issues/94)) and we've not used it internally for years. We provide the instructions to run the Docker image and the ZIP.


colmsnowplow (approved):
LGTM

PR #351, opened 2023-03-20
Correct UMD name for JS tracker snowplow ecommerce plugin

igneel64:


stanch:
@igneel64 I refuse to review this as I have no qualifications or knowledge to do so :)

docs/collecting-data/collecting-from-own-applications/javascript-trackers/javascript-tracker/javascript-tracker-v3/plugins/snowplow-ecommerce/index.md
```diff
@@ -29,7 +29,7 @@ The plugin is available since version 3.8 of the tracker.
 window.snowplow(
     'addPlugin',
     'https://cdn.jsdelivr.net/npm/@snowplow/browser-plugin-snowplow-ecommerce@latest/dist/index.umd.min.js',
-    ['snowplowEcommerceTracking', 'SnowplowEcommercePlugin']
+    ['snowplowEcommerceAccelerator', 'SnowplowEcommercePlugin']
```
matus-tomlein, line 32: Tbh, I would prefer if we didn't call this accelerator since I think the plugin should not be limited to the use in the accelerator, but I see that we already use that identifier in the published package.

matus-tomlein (approved):

docs/collecting-data/collecting-from-own-applications/javascript-trackers/javascript-tracker/javascript-tracker-v3/plugins/snowplow-ecommerce/index.md
igneel64: Yeap this will be changed in V4. In general we need better documentation about these two attributes and how they connect to the whole flow of the plugin addition.

PR #353, opened 2023-03-20
Use variables for tracker versions

mscwilson:
Storing all the tracker versions in one place so they don't get out of date.


stanch (approved):
Excellent!

PR #354, opened 2023-03-20
Fix ruby tracker pages

mscwilson:
The Ruby tracker docs got slightly broken during the Docusaurus migration. I've replaced the broken accordions with tabs. Also added some missing config details.


stanch:
Looks good to me but can’t validate the technical accuracy

PR #357, opened 2023-03-21
Correct and clarify type casting rules for redshift.

voropaevp:


docs/understanding-tracking-design/json-schema-type-casting-rules/index.md
```diff
@@ -43,13 +43,40 @@ OR
 }
 ```
 
-`"null"` is not considered for type casting logic, only for the nullability constraint. Type lookup will continue down the table.
+`"null"` is not considered for type casting logic, only for the nullability constraint. Type lookup will continue down
+the table. `"null"`'s position in a list does not matter. Nullability constraints only apply to the first major of the
```
stanch, line 47: I am not sure I understand what it means that nullability constraints only apply to the first major. Does it mean that adding or removing `null` in minor versions has no effect?

stanch, line 47:
```suggestion
the table. `"null"`'s position in a list (`type` or `enum`) does not matter. Nullability constraints only apply to the first major of the 
```

docs/understanding-tracking-design/json-schema-type-casting-rules/index.md
```diff
@@ -43,13 +43,40 @@ OR
...
+{
+    "properties": {
+      "f1": {"type": "T"}
+    },
+    "required": [ "f1" ]
+}
+```
+
+`"required"` is not considered for type casting logic, only for the nullability constraint. Type lookup will continue down the table.
+Fields that are not listed as `"required"` are nullable. Nullability constraints only apply to the first major of the
```
stanch, line 71: Could you add an explanation of what happens when a field is marked as required but has `null` as one of the types or enum values?

docs/understanding-tracking-design/json-schema-type-casting-rules/index.md
```diff
@@ -81,11 +108,17 @@ For the purposes of this  `number` and `integer` are the same.
...
 }
 ```
 
-- `T1, T2, ..`. only contains `"boolean"` **and** `"integer"`
+OR 
+
+  ```json
+{
+    "type": [ "integer", "boolean"]
+}
```
stanch, line 120: Should `{"type": ["number", "integer"]}` be moved up, e.g. after this line? Otherwise I agree with the feedback that it would be never reached because of `{"type": [T1, T2, ...]}`.

stanch (changes requested):

docs/understanding-tracking-design/json-schema-type-casting-rules/index.md
voropaevp: >  Does it mean that adding or removing null in minor versions has no effect?

Correct. It is impossible to add NOT NULL column to existing table. 
Loosening NULL condition theoretically possible but we do no do that.

voropaevp: it also has affect on validation in enrich, but not for warehouse schemas.

docs/understanding-tracking-design/json-schema-type-casting-rules/index.md
voropaevp: it is above that.

docs/understanding-tracking-design/json-schema-type-casting-rules/index.md
```diff
@@ -45,11 +45,40 @@ OR
 
 `"null"` is not considered for type casting logic, only for the nullability constraint. Type lookup will continue down the table.
 
+`"null"`’s position in a list (`type` or `enum`) does not matter.
+
+Note that each major schema version (`1-0-0`, `2-0-0`, etc) results in a new column (name ending with `_1`, `_2`, etc). Once the loader creates a column for a given schema version as `NULL` or `NOT NULL`, it will never alter the nullability constraint for that column. For example, if a field is nullable in schema version `1-0-0` and not nullable in version `1-0-1`, the column will remain nullable. (In this example, the Enrich application will still validate data according to the schema, accepting `null` values for `1-0-0` and rejecting them for `1-0-1`.)
```
stanch, line 50:
```suggestion
Note that each major schema version (`1-0-0`, `2-0-0`, etc) results in a new column (name ending with `_1`, `_2`, etc). Once the loader creates a column for a given schema version as `NULLABLE` or `NOT NULL`, it will never alter the nullability constraint for that column. For example, if a field is nullable in schema version `1-0-0` and not nullable in version `1-0-1`, the column will remain nullable. (In this example, the Enrich application will still validate data according to the schema, accepting `null` values for `1-0-0` and rejecting them for `1-0-1`.)
```

docs/understanding-tracking-design/json-schema-type-casting-rules/index.md
```diff
@@ -45,11 +45,40 @@ OR
...
+    },
+    "required": [ "f1" ]
+}
+```
+
+`"required"` is not considered for type casting logic, only for the nullability constraint. Type lookup will continue down the table.
+
+Fields that are not listed as `"required"` are nullable. Fields are _also_ nullable when they are listed as `"required"` but have `"null"` in their `type` or `enum` definition. (In the latter case, the Enrich application will still validate that the field is present, even if it’s `null`.)
+
+Note that each major schema version (`1-0-0`, `2-0-0`, etc) results in a new column (name ending with `_1`, `_2`, etc). Once the loader creates a column for a given schema version as `NULL` or `NOT NULL`, it will never alter the nullability constraint for that column. For example, if a field is not required in schema version `1-0-0` and required in version `1-0-1`, the column will remain nullable. (In this example, the Enrich application will still validate data according to the schema, accepting `null` values for `1-0-0` and rejecting them for `1-0-1`.)
```
stanch, line 75:
```suggestion
Note that each major schema version (`1-0-0`, `2-0-0`, etc) results in a new column (name ending with `_1`, `_2`, etc). Once the loader creates a column for a given schema version as `NULLABLE` or `NOT NULL`, it will never alter the nullability constraint for that column. For example, if a field is not required in schema version `1-0-0` and required in version `1-0-1`, the column will remain nullable. (In this example, the Enrich application will still validate data according to the schema, accepting `null` values for `1-0-0` and rejecting them for `1-0-1`.)
```

stanch (approved):

PR #359, opened 2023-03-22
javascript-tracker-v3: more prominent custom build docs

jethron:
Make documentation on building a version of `sp.js` v3 with custom plugins a bit more prominent. This is a small rewrite of the information located in the [migration guide](https://docs.snowplow.io/docs/collecting-data/collecting-from-own-applications/javascript-trackers/javascript-tracker/v2-to-v3-migration-guide/#build-your-own-spjs), but elevated to its own page (since it couldn't actually be done easily in v2 this location isn't an intuitive place to look).

I've also added some info on bundling in external plugins, which isn't covered on the old page.

Applying the diff and trying the example doesn't currently work properly because the [package.json in snowplow-incubator/snowplow-browser-plugin-simple-template](https://github.com/snowplow-incubator/snowplow-browser-plugin-simple-template/blob/main/package.json) is a bit broken at the moment. You can update the reference to the fork in [jethron/snowplow-browser-plugin-simple-template](https://github.com/jethron/snowplow-browser-plugin-simple-template/blob/main/package.json) instead if you want to test. I'll try to get that merged as well so it works.


docs/collecting-data/collecting-from-own-applications/javascript-trackers/javascript-tracker/javascript-tracker-v3/plugins/bundling-a-custom-plugin-selection/index.md
```diff
@@ -0,0 +1,105 @@
...
+```
+
+Once complete (it might take a minute or two), you'll find your brand new `sp.js` at:
+`trackers/javascript-tracker/dist/sp.js` along with a new sourcemap `sp.js.map` which we suggest hosting together for a better developer experience.
+
+## Including Custom Plugins
+The above only works for the official plugins included in the `snowplow-javascript-tracker` repository.
+If you have [developed your own plugins](/docs/collecting-data/collecting-from-own-applications/javascript-trackers/javascript-tracker/javascript-tracker-v3/plugins/creating-your-own-plugins/index.md), or want to include plugins from third parties, the process is similar, but with some additional small code changes:
+
+- Include the plugin as a dependency in `trackers/javascript-tracker/package.json`
```
igneel64, line 36: Do you think a numbered list would be more fitting here ?

stanch:
Not qualified, withdrawing :)

jrpeck1989:
LGTM

docs/collecting-data/collecting-from-own-applications/javascript-trackers/javascript-tracker/javascript-tracker-v3/plugins/bundling-a-custom-plugin-selection/index.md
jethron: Yeah that makes total sense, my bad. :)

PR #379, opened 2023-04-04
Update Unity tracker docs for version 0.7.0

matus-tomlein:
Updates the tracker version and adds information about the new emitter for WebGL.


stanch (approved):
Looks ok to me but not qualified to approve the actual substance

greg-el (approved):
LGTM! One small nitpicky comment, but I'm not too fussed about leaving it as-is if you don't think the change is necessary.

PR #381, opened 2023-04-05
Improve notes and lists in GTM SS components

adatzer:


jethron (approved):
Didn't realise they were all based off each other. Thanks for keeping these in sync!

stanch (approved):

PR #404, opened 2023-04-18
Browser/JS: crossDomainLinker tips

jethron:
- Extract the `crossDomainLinker` information to a reusable component
- Be more explicit that at a minimum you usually want to not decorate the current domain
- Include a tip on hiding the `_sp` param to prevent data contamination when URLs are shared
- Also fixed some instances where the Browser docs were still using the JS tracker APIs that became clear while making these changes

It was tricky to get the slightly different code snippets to work properly with the reusable component (If there's a better way to do that I'm very open to it).

To see my changes to the JS v2 docs I had to fix the rendering issue there (think newer versions of docusaurus are stricter about the "component names must begin with upper case else it's just html" rule). Also tweaked the closing position of some `mdx-code-block` blocks as it gives better syntax highlighting for mdx components when only the imports are included and the actual components are rendered outside the backticks block.

RE: Ticket BCPF-418

### DIFF
--- a/docs/reusable/javascript-tracker-cross-domain/_index.md
+++ b/docs/reusable/javascript-tracker-cross-domain/_index.md
@@ -156,7 +156,7 @@ You can do this using the [History API](https://developer.mozilla.org/en-US/docs
 <>{ props.lang === "browser" && <>
 
 ```javascript
--trackPageView(); // page URL is https://example.com/?example=123&_sp=6de9024e-17b9-4026-bd4d-efec50ae84cb.1680681134458
+trackPageView(); // page URL is https://example.com/?example=123&_sp=6de9024e-17b9-4026-bd4d-efec50ae84cb.1680681134458
 if (/[?&]_sp=/.test(window.location.href)) {
   history.replaceState(history.state, "", window.location.replace(/&?_sp=[^&]+/, "")); // page URL is now https://example.com/?example=123
 }
--- a/docs/reusable/javascript-tracker-cross-domain/_index.md
+++ b/docs/reusable/javascript-tracker-cross-domain/_index.md
@@ -39,12 +39,46 @@ If you want to decorate every link, regardless of its destination:
 Note that the above will decorate “links” which are actually just JavaScript actions (with an `href` of `"javascript:void(0)"`), or links to email addresses or telephone numbers ([`mailto:`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#linking_to_an_email_address) or [`tel:`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#linking_to_telephone_numbers) schemes).
 To exclude these links, check them explicitly:
 
-<>{ props.children[0] }</>
+<>{ props.lang === "browser" && <>
+
+```javascript
+crossDomainLinker(function(linkElement) {
+  return linkElement.href.indexOf('javascript:') < 0;
+});
+```
+</>
+}</>
+<>{ props.lang === "javascript" && <>
+
+```javascript
+snowplow('crossDomainLinker', function(linkElement) {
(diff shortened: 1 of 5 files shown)


docs/reusable/javascript-tracker-cross-domain/_index.md
```diff
@@ -0,0 +1,112 @@
...
+
+Note that the above will decorate “links” which are actually just JavaScript actions (with an `href` of `"javascript:void(0)"`), or links to email addresses or telephone numbers ([`mailto:`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#linking_to_an_email_address) or [`tel:`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#linking_to_telephone_numbers) schemes).
+To exclude these links, check them explicitly:
+
+<>{ props.children[0] }</>
+
+Alternatively, only count links that parse as web URLs by checking the link's [`hostname`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLAnchorElement/hostname).
+This should automatically exclude links that don't lead simply to other web pages.
+
+<>{ props.children[1] }</>
```
stanch, line 47: How does this work when children are in markdown? Each paragraph is a child? Would this break easily if someone unaware of this logic writes an extra paragraph? I’d wrap the text at the call site into two `<div>`s just for good measure.

stanch (commented):
To Ryan’s point, this could also be a dedicated page with tabs for JavaScript tracker and Browser tracker (I think that’s what he meant?)

rlh1994:
>To Ryan’s point, this could also be a dedicated page with tabs for JavaScript tracker and Browser tracker (I think that’s what he meant?)

I meant use JSX (?) to render the value for that language, so like 

```
<CrossDomain lang="javascript">
```
would render the JS snippets, 
```
<CrossDomain lang="browser">
```
would render the browser ones and so on

jethron:
> ```
> <CrossDomain lang="javascript">
> ```
This was actually what I started with, but I couldn't get it to play nice inside the component (especially with the ternary operator, which it tries to render parts of as text for some reason). :sweat_smile:

I've swapped back to something more like that approach, but checking each lang explicitly now so we can just use `&&`. The syntax is _extremely_ fragile: adding/removing a new line in the wrong spot makes it a syntax error, so I'm not a huge fan - but it works for now.

jethron:
> That's just JSX for you, if you don't have new lines after/before some, but not all, tags or blocks it errors in an extremely non-useful error message way

You mean MDX? I don't think I've ever had an issue like this before with JSX in JS.

docs/reusable/javascript-tracker-cross-domain/_index.md
```diff
@@ -0,0 +1,183 @@
...
+
+Because enrichment populates dedicated fields with the parameter, and it doesn't get persisted through the user's session (it is essentially only on the landing page), you can choose to hide the parameter after the first event (e.g. Page view) is fired on the destination page without needing additional modeling to stitch the IDs.
+This means after the next page is tracked, if the user copies the URL after the tracking code has fired, it may not include the `_sp` parameter and get copied to other users they share it with.
+
+You can do this using the [History API](https://developer.mozilla.org/en-US/docs/Web/API/History/replaceState).
+
+<>{ props.lang === "browser" && <>
+
+```javascript
+-trackPageView(); // page URL is https://example.com/?example=123&_sp=6de9024e-17b9-4026-bd4d-efec50ae84cb.1680681134458
```
stanch, line 159:
```suggestion
trackPageView(); // page URL is https://example.com/?example=123&_sp=6de9024e-17b9-4026-bd4d-efec50ae84cb.1680681134458
```

jethron:
Added @igneel64 for review on the tech-side, then. :)

PR #407, opened 2023-04-18
Add accelerator CTA to relevant recipes 

rlh1994:


Jack-Keene:
Could we add a generic CTA for the recipes without a relevant accelerator?

Jack-Keene:
Could we make it so the link opens in a new tab? I think you'd just need to add `target='_blank'` to the <a> tag in https://github.com/snowplow/documentation/blob/main/src/components/AcceleratorAdmonitions/index.js

Jack-Keene (approved):
LGTM

PR #411, opened 2023-04-19
Python 1.0.0 Release

Jack-Keene:


docs/collecting-data/collecting-from-own-applications/python-tracker/previous_versions/python-v0-15/upgrading/index.md
```diff
@@ -0,0 +1,13 @@
+---
+title: "Upgrading to Newer Versions"
+date: "2023-01-25"
+sidebar_position: 100
+---
+
+This page gives instructions to upgrading to newer versions of the Python Tracker
+
+## Upgrading to v0.13.0
```
stanch, line 9: Upgrading to 0.13.0 inside a 0.15 version page? 🤔

Jack-Keene: Just a draft atm! Currently re-writing them all for V1

docs/collecting-data/collecting-from-own-applications/python-tracker/subject/index.md
```diff
@@ -0,0 +1,236 @@
...
+- [`set_color_depth`](#set-color-depth-withset_color_depth)
+- [`set_timezone`](#set-timezone-withset_timezone)
+- [`set_lang`](#set-the-language-withset_lang)
+
+If you initialize a `Tracker` instance without a subject, a default `Subject` instance will be attached to the tracker. You can access that subject like this:
+
+```python
+t = Tracker(my_emitter)
+t.subject.set_platform("mob").set_user_id("user-12345").set_lang("en")
+```
```
matus-tomlein, line 31: Recently we got the feedback from customers that our docs for the .NET tracker don't explain well how to set per-event subject. I think the problem is similar here – event level subject is explained just through a short snippet at the end of the page. Maybe we could explain right at the beginning that there are two ways to set the subject – for the whole tracker or for an event. You can do [something similar as we did in the .NET docs](https://docs.snowplow.io/docs/collecting-data/collecting-from-own-applications/net-tracker/subject/).

docs/collecting-data/collecting-from-own-applications/python-tracker/tracking-specific-events/index.md
```diff
@@ -1,528 +1,282 @@
...
-| `event_json` | The properties of the event | Yes | SelfDescribingJson |
-| `context` | Custom context for the event | No | List(SelfDescribingJson) |
-| `tstamp` | When the unstructured event occurred | No | Positive integer |
-| `event_subject` (since v0.9.0) | The subject for the event | No | Subject |
+| `page_url` | URL of the viewed page | `string` | Yes |
+| `page_title` | Title of the viewed page | `string` | No |
+| `referrer` | The address which linked to the page | `string` | No |
+| `event_subject` | The subject for the event | `Subject` | No |
+| `context` | Custom context for the event | `List(SelfDescribingJson)` | No |
+| `true_timestamp` | When the page view occurred | `int | float` | No |
```
matus-tomlein, line 61: The | in `int | float` is understood by Docusaurus as another column so it's not rendered correctly.

```suggestion
| `true_timestamp` | When the page view occurred | `int or float` | No |
```

docs/collecting-data/collecting-from-own-applications/python-tracker/tracking-specific-events/index.md
```diff
@@ -1,528 +1,282 @@
...
-  {
-    "latitude": -23.2,
-    "longitude": 43.0
-  }
+```python 
+page_view = PageView(
+  page_url="https://www.snowplow.io",
+  page_title="Homepage",
 )
+t.track(page_view)
```
matus-tomlein, line 16: `t` is not defined on this page, so worth spelling it out to `tracker` (better not to use any abbreviations in code especially in the docs snippets).

```suggestion
tracker.track(page_view) 
```

docs/collecting-data/collecting-from-own-applications/python-tracker/tracking-specific-events/index.md
```diff
@@ -1,528 +1,282 @@
...
-        "sku": "pbz0038",
-        "price": 15,
-        "quantity": 1
-    }])
-```
-
-### Track ecommerce transaction items
+- `event_subject`
+- `context`
+- `tstamp`
```
matus-tomlein, line 163:
```suggestion
- `true_timestamp`
```

docs/collecting-data/collecting-from-own-applications/python-tracker/tracking-specific-events/index.md
```diff
@@ -1,528 +1,282 @@
...
-```python
-t.track_form_change("signupForm", "ageInput", "age", "24", "number", ["signup__number", "form__red"])
+t.track(page_view)
 ```
 
-### Track submitted forms
+**Important:** Even if only one custom context is being attached to an event, it still needs to be wrapped in an list.
 
-Use `track_form_submit()` to track sumbitted forms. Arguments are:
+Note also that you should not pass in an empty list of contexts as this will fail validation. Instead of an empty list, you can pass in `None`.
```
matus-tomlein, line 254: Is this really true? That sounds like something that should be handled in the tracker rather than bothering users about it – if the list of context is empty, ignore it.

docs/collecting-data/collecting-from-own-applications/python-tracker/upgrading/index.md
```diff
@@ -1,13 +1,21 @@
...
 sidebar_position: 100
 ---
 
 This page gives instructions to upgrading to newer versions of the Python Tracker
+## Upgrading to v1.0.0
+There are a number of breaking changes and new features in the V1 release, in order to make use of the latest tracker make sure to:
 
+- Tracker `namespace` is now a required property in the tracker and should be parsed on initialization of a new tracker.
+- The `track_xxx()` functions have been deprecated in favor of the `Event` class, further information on tracking events using the new API can be found on the [Event Tracking](collecting-data/collecting-from-own-applications/python-tracker/tracking-specific-events/index.md) page.
+- The `track()` function in the tracker now returns an event id, instead of a tracker instance. Chaining methods using the `track()` function will no longer work, and methods should be applied to the tracker itself.
```
matus-tomlein, line 13: It also doesn't accept a `Payload` instance as was the case before.

docs/collecting-data/collecting-from-own-applications/python-tracker/upgrading/index.md
```diff
@@ -1,13 +1,21 @@
...
-date: "2023-01-25"
+date: "2023-06-14"
 sidebar_position: 100
 ---
 
 This page gives instructions to upgrading to newer versions of the Python Tracker
+## Upgrading to v1.0.0
+There are a number of breaking changes and new features in the V1 release, in order to make use of the latest tracker make sure to:
 
+- Tracker `namespace` is now a required property in the tracker and should be parsed on initialization of a new tracker.
```
matus-tomlein, line 11: We also changed the order of the parameters in the `Tracker` constructor.

docs/collecting-data/collecting-from-own-applications/python-tracker/emitters/index.md
```diff
@@ -38,81 +38,51 @@ def __init__(
...
 | `byte_limit` | Number of bytes to store before flushing | No | Positive integer | `None` |
-| `request_timeout` | Timeout for HTTP requests | No | Positive integer or tuple of 2 integers | `None` | v0.10.0 |
-| `max_retry_delay_seconds` | The maximum time between attempts to send failed events to the collector | No | int | 60s | v0.13.0 |
-| `buffer_capacity` | The maximum capacity of the event buffer | No | int | `None` | v0.13.0 |
-| `custom_retry_codes` | Custom retry rules for HTTP status codes received in emit responses from the Collector | No | dict | `None` | v0.13.0 |
-| `event_store` | Stores the event buffer and buffer capacity | No | EventStore | `None` | v0.13.0 |
-| `session` | Persist parameters across requests by using a session object | No | requests.Session | `None` | v0.15.0 |
+| `request_timeout` | Timeout for HTTP requests | No | Positive integer or tuple of 2 integers | `None` | 
+| `max_retry_delay_seconds` | The maximum time between attempts to send failed events to the collector | No | int | 60s | 
+| `buffer_capacity` | The maximum capacity of the event buffer | No | int | `None` |
```
rlh1994, line 53: default is 60 not 60s

rlh1994, line 53: Is it worth adding that despite a deafult value of None, it is technically 10,000 as noted below?

docs/collecting-data/collecting-from-own-applications/python-tracker/tracking-specific-events/index.md
Jack-Keene: This is something we fixed in V1 so I'll remove this line!

docs/collecting-data/collecting-from-own-applications/python-tracker/previous_versions/python-v0-15/setup/index.md
```diff
@@ -0,0 +1,92 @@
...
+$ pip install snowplow-tracker[celery]
+```
+
+To add the Snowplow Tracker as a dependency to your own Python app, edit your `requirements.txt` and add:
+
+
+<CodeBlock language="bash">{
+`snowplow-tracker==${versions.pythonTracker}`
+}</CodeBlock>
+
```
rlh1994, line 65: This is going to be the newest version, if this is 0-15 you should hardcode it here

docs/collecting-data/collecting-from-own-applications/python-tracker/previous_versions/python-v0-15/upgrading/index.md
rlh1994: This comment is still relevant :D

docs/collecting-data/collecting-from-own-applications/python-tracker/subject/index.md
```diff
@@ -0,0 +1,236 @@
...
+
+We will discuss each of these in turn below:
+
+### Change the tracker's platform with `set_platform`
+
+The default platform is "pc". You can change the platform the subject is using by calling for example:
+
```
rlh1994, line 38: would prefer this be backticked rather than quoted but not that bothered

docs/collecting-data/collecting-from-own-applications/python-tracker/subject/index.md
```diff
@@ -0,0 +1,236 @@
...
+```
+
+### Set screen resolution with `set_screen_resolution`
+
+If your Python code has access to the device's screen resolution, then you can pass this in to Snowplow too. 
+Both numbers should be positive integers; note the order is width followed by height. For example:
+
+```python
+s.set_screen_resolution(1366, 768)
+```
```
rlh1994, line 60: This doesn't need changing, but what a cursed screen resolution

docs/collecting-data/collecting-from-own-applications/python-tracker/subject/index.md
```diff
@@ -0,0 +1,236 @@
...
+
+### Set viewport dimensions with `set_viewport`
+
+If your Python code has access to the viewport dimensions, then you can pass this in to the subject too. Both numbers should be positive integers; note the order is width followed by height. For example:
+
+```python
```
rlh1994, line 66: These are basically the same paragraph but in one you say it can be passed to the subject, the other says to Snowplow, can we be consistent throughout the whole page please

rlh1994: To be honest most of these say the same thing of "If you have access to X, you can set it with Y, it should be of type Z". Is there a better way to rework this to reduce duplication?

docs/collecting-data/collecting-from-own-applications/python-tracker/subject/index.md
```diff
@@ -0,0 +1,236 @@
...
+The `domain_userid` field of the Snowplow event model corresponds to the ID stored in the first party cookie set by the Snowplow JavaScript Tracker. If you want to match up server-side events with client-side events, you can set the domain user ID for server-side events like this:
+
+```python
+s.set_domain_user_id('c7aadf5c60a5dff9')
+```
+
+You can extract the domain user ID from the cookies of a request using the `get_domain_user_id` function below. The `request` argument is, as an example, the [Django request object](https://docs.djangoproject.com/en/1.7/ref/request-response/).
+
+**Note that this function has not been tested.**
+
```
rlh1994, line 121: can we test it?

docs/collecting-data/collecting-from-own-applications/python-tracker/subject/index.md
```diff
@@ -0,0 +1,236 @@
...
+    return None
+
+def get_domain_user_id(request):
+    cookie = snowplow_cookie(request)
+    if cookie != None:
+        return cookie.split(".")[0]
+```
+
+If you used the "cookieName" configuration option of the Snowplow JavaScript Tracker, replace "_sp_" with the same string you passed as the cookieName.
+
```
rlh1994, line 137: again quotes mixed in

docs/collecting-data/collecting-from-own-applications/python-tracker/tracking-specific-events/index.md
```diff
@@ -1,528 +1,282 @@
...
-
-```python
-t.track_struct_event("some cat", "save action", tstamp=1368725287000)
-```
-
-**_\*\*Prior to v0.9.0_**
+Every tracked event payload has a unique event_id UUID string (`eid`). Other properties include the name_tracker (`namespace`) and `app_id` set when the Tracker was initialized. From version 1 onwards, `tracker.track()` returns the payload's `eid`.
 
-Before version 0.9.0 of the Python Tracker, providing a `snowplow_tracker.timestamp.TrueTimestamp` object as the timestamp argument will attach a true timestamp to the event, replacing the device timestamp. For example:
+Snowplow events have a defined structure and protocol that is identical regardless of the tracker used. A minimal payload - the raw event - is sent from the tracker to your collector. The raw event is [enriched](/enriching-your-data/what-is-enrichment/index.md) as it passes through your pipeline. By the time the event arrives in your data storage, depending which [enrichments](/enriching-your-data/available-enrichments/index.md) you have enabled, it will have gained different kinds of metadata, and have many more fields than it started with.
```
rlh1994, line 22: This is nice but feels like we shouldn't explaining the whole pipeline in the docs of one of the trackers? Also use full links if you do keep it.

docs/collecting-data/collecting-from-own-applications/python-tracker/tracking-specific-events/index.md
```diff
@@ -1,528 +1,282 @@
...
+The Python tracker provides classes for tracking different types of events. They are listed below:
 
-```python
-
-evSubject = Subject().set_user_id("1234")
-t.track_page_view("www.example.com", event_subject=evSubject)
-```
+| **Event** | **Description** |
+| --- | --- |
+| `SelfDescribing` | Track custom events with custom schemas |
```
rlh1994, line 30: I'd argue that SelfDescribing are not really out of the box events, they aren't classed as such in other areas on the docs. I would maybe split this one out and add the section about it beneath all the others as it's a more complex use case.

docs/collecting-data/collecting-from-own-applications/python-tracker/tracking-specific-events/index.md
```diff
@@ -1,528 +1,282 @@
...
+  referrer="https://docs.snowplow.io/docs",
+)
+t.track(page_ping)
 ```
 
-### Track page pings
+## Creating a `ScreenView` Event
 
-Use `track_page_ping()` to track engagement with a web page over time, via a heartbeat event. (Each ping represents a single heartbeat.)
+Use the `ScreenView` to track a user viewing a screen (or equivalent) within your app.
```
rlh1994, line 101: When should I use screen view and when should I use page view?

docs/collecting-data/collecting-from-own-applications/python-tracker/tracking-specific-events/index.md
```diff
@@ -1,528 +1,282 @@
...
-| `id` | Order ID | Yes | Non-empty string |
-| `sku` | Item SKU | Yes | Non-empty string |
-| `price` | Item price | Yes | Int or Float |
-| `quantity` | Item quantity | Yes | Int |
-| `name` | Item name | No | String |
-| `category` | Item category | No | String |
-| `context` | Custom context for the event | No | List(SelfDescribingJson) |
-| `tstamp` | When the transaction event occurred | No | Positive integer |
-| `event_subject` (since v0.9.0) | The subject for the event | No | Subject |
+Event level subjects are combined with any tracker subjects that have been initialized, with the event subject taking priority over tracker subject parameters.
```
rlh1994, line 169: Link to the subject page here?

docs/collecting-data/collecting-from-own-applications/python-tracker/tracking-specific-events/index.md
```diff
@@ -1,528 +1,282 @@
...
-```
+Whilst each of those events is a different type, all of them involve capturing data about the user and the video. Both the 'user' and 'video' are entities that are tracked across multiple event types. Both are candidates to be "custom context". You as a Snowplow user can define your own custom contexts (including associated schemas) and then send data for as many custom contexts as you wish with _any_ Snowplow event. So if you want, you can define your own "user context", and then send additional user data in that object with any event. Other examples of context include:
 
-### Track add-to-cart events
+- articles
+- videos
+- products
+- categories
+- pages / page_types
+- environments
```
rlh1994, line 207: Again this all feels like it's already explained elsewhere in the docs, link to the docs for contexts and keep this page just for what's specific to the python tracker?

docs/collecting-data/collecting-from-own-applications/python-tracker/tracking-specific-events/index.md
```diff
@@ -1,528 +1,282 @@
...
+  context=[poster_context, geo_context]
+)
 
-```python
-t.track_form_change("signupForm", "ageInput", "age", "24", "number", ["signup__number", "form__red"])
+t.track(page_view)
 ```
 
-### Track submitted forms
+**Important:** Even if only one custom context is being attached to an event, it still needs to be wrapped in an list.
```
rlh1994, line 252: Picky, but more that the object needs to be a list, rather than wrapped in.

rlh1994 (changes requested):
A few nitpick bits, and a few bad links that need changing. I haven't much gone over the previous version stuff because I've just assumed that's fine. And also I haven't validated the actual _code_ for the most part.

docs/collecting-data/collecting-from-own-applications/python-tracker/tracking-specific-events/index.md
```diff
@@ -1,528 +1,282 @@
...
+| `PagePing` | Track engagement on web pages over time |
+| `ScreenView` | Track views of a screen |
+| `StructuredEvent` | Track custom events without schemas |
 
-### Tracker method return values
+## Creating a custom event (`SelfDescribing`)
+To track data using a `SelfDescribing` event, the data must be structured as a `SelfDescribingJson` object. These require two fields, a URI for a self-describing JSON schema and the data in the form of a `PayloadDict`. The data must be valid against the schema. 
 
-All tracker methods will return the tracker instance, allowing tracker methods to be chained:
+A simple initialisation for a link click event looks like this:
```
rlh1994, line 39: American spelling please

docs/collecting-data/collecting-from-own-applications/python-tracker/tracking-specific-events/index.md
```diff
@@ -1,528 +1,282 @@
...
-| `filters` | Filters applied to search | No | List(dict{str:str |
-| `total_results` | Total number of results | No | Integer |
-| `page_results` | Number of pages of results | No | Integer |
-| `context` | Custom context for the event | No | List(SelfDescribingJson) |
-| `tstamp` | When the pageview occurred | No | Positive integer |
-| `event_subject` (since v0.9.0) | The subject for the event | No | Subject |
+- the `dvce_sent_tstamp`, which is the timestamp when the event was sent
 
-Basic example:
+These are going to be used downstream, to calculate the `derived_tstamp` for the event, which takes also into account the collector timestamp, in order to best approximate the exact time the event occured.
```
rlh1994, line 266: *occurred

docs/collecting-data/collecting-from-own-applications/python-tracker/emitters/index.md
```diff
@@ -123,79 +93,35 @@ def failure(num, arr):
...
 
-:::note New in v0.13.0
-
-Since version 0.13.0, the constructor takes another `event_store` argument.
-
-:::
-- `event_store`
+### `event_store`
 
 The event store is used to store an event queue with events scheduled to be sent. Events are added to the event store when they are tracked and removed when they are successfuly emitted or when emitting fails without any scheduled retries. The default is an InMemoryEventStore object with a buffer_capacity of 10,000 events.
```
rlh1994, line 119: successfully

rlh1994 (changes requested):
words

docs/collecting-data/collecting-from-own-applications/python-tracker/subject/index.md
Jack-Keene: Apparently this is a 720p screen!

docs/collecting-data/collecting-from-own-applications/python-tracker/tracking-specific-events/index.md
Jack-Keene: Screenview is the app equivalent of pageview for the web, I've clarified in the pageview section and the table summary

docs/collecting-data/collecting-from-own-applications/python-tracker/subject/index.md
Jack-Keene: I'm just going to bin it off, not sure we really need django docs in here.

rlh1994: unchained, but i'll allow it

matus-tomlein (approved):

PR #425, opened 2023-04-27
Add conversions information - web 0.15.0

rlh1994:
Add information for the upcoming conversions stuff, and bump the version number.

I've provided some examples for the configuration, but I've done this from a what fields you're using point of view, rather than a what you are trying to achieve POV (e.g. the ecommerce one is highlighted as using a sde and context field) because this way they aren't limited to a use case and instead cover a wider range and are more likely to be find what they are looking for, but I'm happy to change this and feel free to suggest any more to add!


docs/modeling-your-data/modeling-your-data-with-dbt/dbt-models/dbt-web-data-model/conversions/index.md
```diff
@@ -0,0 +1,144 @@
...
+
+| Column                       | Description                                                                                  |
+| ---------------------------- | -------------------------------------------------------------------------------------------- |
+| `cv_{name}_volume`           | An integer column with the total number of conversions in the session                        |
+| `cv_{name}_values`           | An array of values for each conversion, if provided in the configuration                     |
+| `cv_{name}_total`            | The total of the values for each conversion, if provided in the configuration                |
+| `cv_{name}_events`           | An array of `event_id`s for each conversion, if `list_events` is `true` in the configuration |
+| `cv_{name}_first_conversion` | The `collector_tstamp` of the first conversion event                                         |
+| `cv_{name}_converted`        | A boolean for if there was a conversion in the session                                       |
```
miike, line 37: I think we'll also need a cv_summary column (this makes it more straightforward for us to use in dashboards as the column name won't vary) for total value (sum) and count of total conversions.

rlh1994: So a total sum and count across all the listed conversions?

docs/modeling-your-data/modeling-your-data-with-dbt/dbt-models/dbt-web-data-model/conversions/index.md
```diff
@@ -0,0 +1,204 @@
...
+
+| Column                       | Description                                                                                  |
+| ---------------------------- | -------------------------------------------------------------------------------------------- |
+| `cv_{name}_volume`           | An integer column with the total number of conversions in the session                        |
+| `cv_{name}_values`           | An array of values for each conversion, if provided in the configuration                     |
+| `cv_{name}_total`            | The total of the values for each conversion, if provided in the configuration                |
+| `cv_{name}_events`           | An array of `event_id`s for each conversion, if `list_events` is `true` in the configuration |
+| `cv_{name}_first_conversion` | The `collector_tstamp` of the first conversion event                                         |
```
miike, line 77: We should probably change this to `derived_tstamp` so that we can allow for historical conversions to be sent in.

rlh1994:
So there's a very weird reproducible bug in this PR, but it _only_ happens locally - if you go start the docs locally, then go to the web config page (http://localhost:3000/docs/modeling-your-data/modeling-your-data-with-dbt/dbt-configuration/web/) you can change the theme quickly and without issue, if you click any header link in the RHS sidebar of this page, then try to change the theme, it crashes the entire page. It doesn't happen on the deployed preview, and also doesn't happen when you go directly to a header via a url, it only happens AFTER you click a header link...

agnessnowplow (approved):

PR #427, opened 2023-05-02
Update iOS tracker version to 5.0.1 and add cards for installation page

matus-tomlein:
1. updates iOS tracker version to 5.0.1
2. adds cards to empty page for installation and setup on mobile tracekrs


rlh1994 (approved):
This must have taken you forever!

PR #446, opened 2023-05-18
Feat/runnable snippets

greg-el:
This PR adds the ability for users of the docs to run a limited number of javascript code snippets in the docs. The user can set their own Application ID and Collector URL. 

This is currenly only on this page:
https://deploy-preview-446--snowplow-docs.netlify.app/docs/collecting-data/collecting-from-own-applications/javascript-trackers/javascript-tracker/javascript-tracker-v3/tracking-events/

Note that the Collector URL is limited to either https or http localhost, as we cannot send events from a secure site (the docs site) to a http collector.

## UI Changes

What were previously hover buttons (copy/wrapping) are now contained on a bar on top of each snippet, to make any "runnable" snippets obvious  without hovering (shown as ones with the "play" button on the left side of the code block top bar. Top snippet is `runnable`, bottom is not.):

image

## How it works

This feature is configured by accessing a menu from a new navbar item, next to the theme icon, which shows only on pages with runnable snippets.
image

This opens a side menu which allows the setting of a collector URL and app id. The side menu will also open if a user clicks on a "run" button without having the feature enabled.

<details>
<summary><strong>The side menu</strong>
</summary>
image
</details>

The navbar icon when the feature is enabled:
image

## For docs writers

For the code snippet blocks, this feature is enabled on snippets with a new metatag, `runnable`. This can be used as such for regular, triple backtick, code snippets:

\```javascript **runnable**
snowplow('trackPageView');
\```

Or as a `CodeBlock`:

```
<CodeBlock language="javascript" runnable>{
    `snowplow('trackPageView');
`}</CodeBlock>
```

Once this tag has been added, and the feature enabled, the "run" button will execute the contents of the code block.


matus-tomlein (commented):
This is really cool! And a smart solution with using the iframe as a sandbox. Too bad this won't work with the browser tracker snippets but that's probably a separate task to address that.

Just a couple of things that I noticed trying this out:

1. When you resize the window width to be smaller such that the sidebar goes away, the play button is no longer visible on the snippets – is this intentional?

2. It works for me on Chrome but I can't configure the collector URL on Safari, I get this in the logs:

```
Not allowed to request resource
Fetch API cannot load http://localhost:9090/i?e=pv&aid=test due to access control checks.
```

I think it's a problem with the CORS policy on the docs website. I get a similar problem when trying this in the console on Chrome:

```
> await fetch("http://localhost:9090/health")
deploy-preview-446--snowplow-docs.netlify.app/:1 Access to fetch at 'http://localhost:9090/health' from origin 'https://deploy-preview-446--snowplow-docs.netlify.app' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```

I think we might run into a similar problem when the website is deployed on docs.snowplow.io. When trying to run the fetch from localhost in a console there, I get the same CORS error.

Do we need to set the `Access-Control-Allow-Origin` header to allow localhost calls on the docs website? I am a bit confused why this works on Chrome while running the fetch from a console in there gives the error...

greg-el:
After chatting to Nick on slack, it seems that `no-cors` doesn't solve the issue on Safari. I haven't been able to come up with a solution, so may just require a warning that it doesn't work on Safari.

@matus-tomlein 
1. Yeah that was intentional, as we decided not to bother supporting mobile due to the very low amount of users that we see use the docs on mobile, so the UI elements hide themselves when the width what docusaurus considers to be mobile.

Jack-Keene:
Just a few things I noticed when testing this out:

1. When I test the page ping events/enable activity tracking, I don't get a page ping event sent (the page view at the bottom of the block does get sent). 
2. Would it be useful to strip trailing slashes from the collector string before it sends a request the health check endpoint? 
3. Not sure if this is expected behaviour, but when I change the collector/app id in the drawer, a page view event gets sent to that collector.

greg-el:
1. It seems the reason for that is because the "page" is actually the iframe the tracker sits in, and as there isn't any interaction within the iframe, the event is never fired. I've removed the `trackable` metatag from that example.
2. Yep, fixed
3. Yeah that is expected, we use a "fake" page view to test if the collector is reachable, if `/health` isn't avilable.

Jack-Keene (approved):
LGTM!

PR #448, opened 2023-05-19
Add privs block to BDP page

rlh1994:
Had a few support tickets that stem back to this issue in the last few weeks, getting ahead of it.


georgewoodhead (approved):
LGTM!

PR #450, opened 2023-05-23
JS Trackers: More DocCardLists

jethron:
Per the repo analytics dashboard's "Collecting Data" feedback:

> [Where’s the v2 docs?](https://docs.snowplow.io/docs/collecting-data/collecting-from-own-applications/javascript-trackers/javascript-tracker/javascript-tracker-v2/)

> [API docs](https://docs.snowplow.io/docs/collecting-data/collecting-from-own-applications/javascript-trackers/browser-tracker/browser-tracker-v3-reference/)

I'm assuming this is because those pages are basically stubs and it's not obvious you need to check the navigation on the left to see the content. The existing DocCardLists component adds tiles and is a good fit for these otherwise pretty empty pages so I've added it there.

This PR:
- adds cards to the pages in question linking to their categories' pages, which include links to the generated API docs
- adds cards to other "stub" pages in those sections where relevant
- move the tracker v3 announcement present on most tracker v2 pages to a reusable block, and isolated it in an admonition so it's more clearly not a part of the actual v2 doc content and is easier to update when v4 drops
- fix a random path issue that brakes the build on my local for some reason?

Also while making these changes I notice that the v2 tracker is still marked ["Actively Maintained"](https://docs.snowplow.io/docs/collecting-data/collecting-from-own-applications/javascript-trackers/javascript-tracker/javascript-tracker-v2/), @igneel64 should we consider downgrading this to just "Maintained" per [the classification](https://docs.snowplow.io/docs/collecting-data/collecting-from-own-applications/tracker-maintenance-classification/) ?

### DIFF
--- a/docs/collecting-data/collecting-from-own-applications/javascript-trackers/browser-tracker/browser-tracker-v3-reference/advanced-usage/index.md
+++ b/docs/collecting-data/collecting-from-own-applications/javascript-trackers/browser-tracker/browser-tracker-v3-reference/advanced-usage/index.md
@@ -5,3 +5,9 @@ sidebar_position: 4000
 ---
 
 This section covers the more advanced elements of the JavaScript Tracker.
+
+```mdx-code-block
+import DocCardList from '@theme/DocCardList';
+```
+
+<DocCardList/>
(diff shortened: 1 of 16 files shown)


rlh1994 (approved):
LGTM!

jethron:
All good, thanks @emielver!

PR #454, opened 2023-05-23
Update docs for gtm-ss LaunchDarkly tag 0.2.0

adatzer:


jethron (approved):
LGTM, can we update the UA example under `Versioning > Version` too? (Still mentions "CompanyName").

PR #458, opened 2023-05-26
Update docs for C++ tracker version 2

matus-tomlein:
Update installation instructions with the new install option in v2. Add a section to the migration guide.

### DIFF
--- a/docs/collecting-data/collecting-from-own-applications/c-tracker/setup/index.md
+++ b/docs/collecting-data/collecting-from-own-applications/c-tracker/setup/index.md
@@ -21,17 +21,41 @@ Supported operating systems:
 
 ## Installing in your project
 
-There are two ways to install the tracker in your app:
+There are three ways to install the tracker in your app:
 
-1. By including the project using cmake.
-2. By copying source files inside the `include` folder into your codebase.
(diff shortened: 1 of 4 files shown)


docs/collecting-data/collecting-from-own-applications/c-tracker/setup/index.md
```diff
@@ -21,26 +21,49 @@ Supported operating systems:
 
 ## Installing in your project
 
-There are two ways to install the tracker in your app:
+There are three ways to install the tracker in your app:
 
-1. By including the project using cmake.
-2. By copying source files inside the `include` folder into your codebase.
+1. By adding the project into your `CMakeLists.txt` as a subdirectory.
```
greg-el, line 26: These don't seem to be in the same order as the actual ways below, it would be nice if they matched up

greg-el (approved):
LGTM, just a small comment about the list order

docs/collecting-data/collecting-from-own-applications/c-tracker/setup/index.md
matus-tomlein: hmm, they should be: 1. subdirectory, 2. import target, 3. copy files.

PR #459, opened 2023-05-26
Add information to .NET tracker docs how to configure global and event subject

matus-tomlein:
Adds missing code snippets showing how to set the subject information on the tracker as well as on the event level in the .NET tracker.


mscwilson (approved):
lgtm

PR #467, opened 2023-06-02
Highlight offerings

stanch:
This PR adds a mechanism to automatically highlight different offerings (BDP Enterprise, BDP Cloud, Open Source) and updates a couple of pages to demo this. Updating the remaining pages will come in a different PR.

# Guiding principles

* Make it clear what is in the paid product and what is in the free product
* Help Open Source users discover compelling paid functionality
* Help customers skip what is only relevant to Open Source users
* Automatic signposting in the sidebar, cards _and_ each offering-specific page (including subpages)
* Make it possible to add an offering selector that would filter the sidebar (future work)
* Make it possible to integrate with search facets (future work)
* Make it possible to discard Open Source specific content in Zendesk Federated search (future work)

# Information architecture

There were a few options:
* ~Completely separate sidebars for Enterprise, Cloud, Open Source~  — does not help Open Source users discover BDP functionality. Also it would be a lot of effort to reuse common pages, which is the majority of the content.
* ~Within each task (e.g. “Managing data structures”), add subcategories for Enterprise, Cloud, Open Source~ — for content that is only in one offering, e.g. Tracking Catalog, this approach awkwardly adds unnecessary dropdowns in the sidebar.
* **Keep existing structure but label pages / categories with an icon** — lightweight, flexible and achieves all goals

# Propagating the offering for a page

There were a few options:
* ~Use directory structure or file names (e.g. `/enterprise/...` or `/enterprise--...`)~ — discarded because we would need to change URLs every time we adjust what is available where
* **Use `sidebar_custom_props`** — clean, achieves all goals (we might still want to change the URLs for open source content to exclude it from Zendesk, but that can be always done in the future)

# Styling the sidebar

There were a few options:
* ~Swizzle the sidebar item~ — unsafe and generated an awful lot of code which would likely create problems when we upgrade Docusaurus
* ~Add the icons to the label in `sidebars.js`~ — workable, but still does not allow to add a tooltip
* **Add icons and tooltips fully via CSS** — a bit meh but achieves all goals

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -244,6 +244,84 @@ table h3 {
   line-height: 20px;
 }
 
+li.enterprise > div > a::before,
+li.enterprise > a::before,
+li.cloud > div > a::before,
+li.cloud > a::before,
+li.opensource > div > a::before,
+li.opensource > a::before {
+  display: inline-block;
+  vertical-align: middle;
+  margin-right: 0.5em;
+}
+
+article.enterprise > a > p::before,
+article.cloud > a > p::before,
+article.opensource > a > p::before {
+  display: inline-block;
+  vertical-align: bottom;
+  margin-right: 0.3em;
+}
+
+li.enterprise > div > a::before,
+li.enterprise > a::before,
+article.enterprise > a > p::before {
+  content: '🏭';
+}
+li.cloud > div > a::before,
+li.cloud > a::before,
+article.cloud > a > p::before {
+  content: '☁️';
+}
+li.enterprise.cloud > div > a::before,
+li.enterprise.cloud > a::before,
+article.enterprise.cloud > a > p::before {
+  content: '🏭 ☁️';
+}
+li.opensource > div > a::before,
+li.opensource > a::before,
+article.opensource > a > p::before {
+  content: '🛠️';
+}
+
+a.menu__link, a.card {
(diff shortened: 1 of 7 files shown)


src/css/custom.css
```diff
@@ -244,6 +244,84 @@ table h3 {
...
+  bottom: auto;
+}
+.enterprise a:hover::after {
+  content: '🏭: For Snowplow BDP Enterprise customers';
+}
+.cloud a:hover::after {
+  content: '☁️: For Snowplow BDP Cloud customers';
+}
+.enterprise.cloud a:hover::after {
+  content: '🏭 ☁️: For Snowplow BDP (Enterprise and Cloud) customers';
```
mscwilson, line 319: This comment didn't fit into the sidebar when I tested it. 
Screenshot 2023-06-06 at 14 07 03

mscwilson (approved):
LGTM! Pretty cool

src/theme/DocCardList/index.js
rlh1994, line 1: If you're already going to swizzle this, could you add the page description to the card when there is one? Currently I don't even know where that field gets used but when I tried before I couldn't managed it...

stanch: Will send in a separate PR

PR #481, opened 2023-06-13
Add migration guide for sql runner 

rlh1994:
Added a page to help users migrate away from SQL Runner to dbt when they are using the default models. Includes a mapping from their current variables to dbt ones and merge statements for each of the tables. Also added a note to the pages about they should use dbt instead if they are starting out.

@stanch because I had to edit the css for all codeblocks on the site because these ones are just really long, I think this is a good thing anyway but I picked the number somewhat arbitrarily.

### DIFF
--- a/docs/modeling-your-data/modeling-your-data-with-sql-runner/migrating-to-dbt/index.md
+++ b/docs/modeling-your-data/modeling-your-data-with-sql-runner/migrating-to-dbt/index.md
@@ -1,0 +1,1137 @@
+---
+title: "Migrating from SQL Runner to dbt"
+sidebar_position: 0
+description: SQL Runner is no longer actively developed and users should try to migrate to dbt where possible, this guide helps you do that
+---
+
+```mdx-code-block
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+```
+
+:::info
+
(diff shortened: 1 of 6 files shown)


docs/modeling-your-data/modeling-your-data-with-sql-runner/migrating-to-dbt/index.md
```diff
@@ -0,0 +1,1137 @@
...
+
+:::
+
+## Why Migrate?
+
+SQL Runner is currently in maintenance mode, while we will continue to fix bugs when they are identified, we are not actively developing the tool or the models anymore and at some point in the future may deprecate it entirely. Our dbt models on the other hand are under active development, with new features and optimizations being made regularly. It is also a more widely used tool and we have a far wider range of packages available in dbt. 
+
+In general, if you are happy with SQL Runner and don't foresee a need to add more models in the future then there is no need to migrate; however if you are starting from scratch, or would like to make use of our wider range of models, then you should consider migrating to dbt.
+
+## Differences between the tools
```
agnessnowplow, line 24: Was thinking we should try and make it more sales pitchy, highlight it more how much more convenient it will be to run things on dbt. Perhaps even cost saving? Not sure if the extra table lookback value (don't ask which one but you will know) for the incremental updates helps in that regard.

rlh1994: I've put a bit more in, but not sure if there's more to add?

agnessnowplow (approved):

PR #493, opened 2023-06-21
Remove the reference to iglu-example-schema-registry

stanch:


colmsnowplow (approved):
LGTM! If we want a full example to point to, we could still point to iglu central, but that might be confusing.

PR #497, opened 2023-06-22
Centralize Snowplow 101 pages

stanch:


stanch:
Hey @jethron and @rlh1994, sorry to pull you in, but you seem like the kind of eagle-eyed people who could review this! I’ve checked this a few times myself, so mainly looking for the following:
* Do the new event, entity and schema pages make sense?
* Was I careful enough in using the correct links everywhere, since I split the event and entity page? In the README, I outlined the rules I tried to follow. (Yes, I know many tracker pages still don’t follow these rules. I will chase some people for that.)

README.md
```diff
@@ -73,6 +73,25 @@ Whenever the same functionality can be achieved in multiple offerings but in a d
...
+There are several key concepts at Snowplow: events (self-describing, structured), entities, schemas. We must ensure that we use and explain them consistently.
+
+**Please, use up-to-date terms:**
+* “Self-describing event”, not “unstructured event”
+* “Entities”, not “contexts”
+* If you are writing about schemas, pick “schema” or “data structure” and stick with it
+
+**Please, do not over-explain these in any of your writing.** Instead, just link to one of the existing concept pages:
+* Events in general: `/docs/understanding-your-pipeline/events/index.md`
+* Custom events: `/docs/understanding-your-pipeline/events/index.md#self-describing-events`
```
stanch, line 87: The link is not a copy-paste error. In most cases where custom events are mentioned, the context (pun not intended) is that schemas are required. And also we recommend these over structured events.

README.md
```diff
@@ -73,6 +73,25 @@ Whenever the same functionality can be achieved in multiple offerings but in a d
...
 
+### Concepts
+
+There are several key concepts at Snowplow: events (self-describing, structured), entities, schemas. We must ensure that we use and explain them consistently.
+
+**Please, use up-to-date terms:**
+* “Self-describing event”, not “unstructured event”
+* “Entities”, not “contexts”
+* If you are writing about schemas, pick “schema” or “data structure” and stick with it
+
```
rlh1994, line 84: Can we also agree how to refer to the entity that is the collection of schema-defined-objects, and the entity that is an individual object defined by the schema?

rlh1994: entity (single object) > entities (list of object) > entitieses (multiple different lists of objects) 🙃

stanch: Hmmmmm, is that a thing (the list of lists)? Where is it found?

stanch: entity, entities, sets of entities? But where does the need for such terms arise?

rlh1994: I'd say we actually use entity, entity, then entities currently. Because we say the YAUAA entity, in e-commerce we use the product entity (but it can be multiple sent at the same time), but you can attach multiple entities to the same event... Feels like we don't have a way to distinguish between the object itself and the list of objects that end up in the same row+column

rlh1994: This was why I was considering context for the list of entities, but I think it's because I'm just used to working with the columns named `context_...`

docs/understanding-your-pipeline/failed-events/index.md
```diff
@@ -24,7 +25,9 @@ When an error occurs while enriching an event (that is part of a collector paylo
```
rlh1994, line 25: The note on line 19 is a little confusing, could you make it a bit clearer?

rlh1994 (commented):
Very nice!

docs/understanding-your-pipeline/architecture-overview-aws/index.md
```diff
@@ -10,28 +10,19 @@ sidebar_position: 0
...
-- Loaders then load your data off the real-time streams into the various Destinations you have set up for your pipeline. Typically these include a Data Warehouse (Redshift, Snowflake, BigQuery) but can also include [many other destinations](/docs/pipeline-components-and-applications/loaders-storage-targets/index.md).
+- The **Enrich** application cleanses the data and validates each event against its [schema](/docs/understanding-your-pipeline/schemas/index.md) to ensure it meets the criteria you have designed and set. When an event fails to validate, it will feed into a bad data stream which contains all of your [failed events](/docs/understanding-your-pipeline/failed-events/index.md). This way, the Snowplow pipeline is non-lossy as failed events [can be reprocessed](/docs/managing-data-quality/recovering-failed-events/index.md).
```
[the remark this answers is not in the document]
stanch: I think I will leave this for some other time. After all it’s just how things are :)

docs/understanding-your-pipeline/events/index.md
```diff
@@ -0,0 +1,159 @@
...
+
+This includes all the [baked-in events](#baked-in-events) and also some [self-describing events](#self-describing-events) that were predefined by the Snowplow team, e.g. link clicks. Tracking SDKs usually provide dedicated API to create out-of-the-box events.
+
+You can also create your own custom events to match your business logic. You can choose between [self-describing events](#self-describing-events) (recommended) and [structured events](#structured-events).
+
+| Out-of-the-box events | Custom events |
+|:-:|:-:|
+| Baked-in events | Structured events |
+| Self-describing events <br/> _(predefined by Snowplow)_ | Self-describing events <br/> _(defined by you)_ |
+
```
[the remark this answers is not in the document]
stanch: Tried to reword the text above, is it better?

stanch:
Rebased and addressed (most) comments

docs/understanding-your-pipeline/events/index.md
```diff
@@ -0,0 +1,167 @@
+---
+title: "Understanding events"
+sidebar_label: "Events"
+sidebar_position: 1
+description: "An event is a central concept in Snowplow that represents something that occurred in a particular point in time"
```
jethron, line 5: Nit: "at a particular point in time", also below.

docs/understanding-your-pipeline/events/index.md
```diff
@@ -0,0 +1,167 @@
...
+* Page views and screen views
+* Page pings
+* Link clicks
+* Form fill-ins (for the web)
+* Form submissions
+* Transactions
+
+Some of these are [baked-in events](#baked-in-events), while others are [self-describing events](#self-describing-events) that were predefined by the Snowplow team, e.g. link clicks. Tracking SDKs usually provide dedicated API to create out-of-the-box events (regardless of their kind).
+
+You can also create custom events to match your business logic. For that purpose, you can either define your own [self-describing events](#self-describing-events) (recommended), or use [structured events](#structured-events).
```
jethron, line 40: Maybe "business requirements" or "own requirements" over "business logic"? I feel like "business logic" implies some kind of action/process that event definitions aren't really related to.

docs/understanding-your-pipeline/failed-events/index.md
```diff
@@ -1,7 +1,8 @@
 ---
 title: "Understanding failed events"
-date: "2020-02-15"
-sidebar_position: 1
+sidebar_label: "Failed events"
+sidebar_position: 4
+description: "Failed events represent data that did not pass validation or otherwise failed to be processed"
```
jethron, line 5: Should we have another "Terminology" block here referencing "bad rows"?

Should the reference/link to recovery be a bit more prominent? (Hidden in a `<details>` atm I think?)

docs/understanding-your-pipeline/failed-events/index.md
```diff
@@ -16,15 +17,19 @@ All failed events are routed to storage (AWS S3 or GCP cloud storage).
 
 :::note
 
-When an error occurs while enriching an event (that is part of a collector payload with several events), the Failed Event will contain **only** this event. The other events of the collector payload can be successfully enriched.
+Collector might receive events in batches. So if something is wrong with the Collector payload as a whole, the generated failed event might represent several Snowplow events.
```
jethron, line 20: Should this be "Collectors might" or "The collector might"?

Also we title case Collector here in the new version, but not earlier in the page?

jethron, line 20: > So if something is wrong with the Collector payload as a whole, the generated failed event might represent several Snowplow events.

I'd probably drop the "So" and say "the entire batch of Snowplow events" rather than just "several" which makes it sound like some may still be OK.
Should we give an example of some of the failure types that would impact the whole batch vs individual events? (Collector Payload Format Violation, Adaptor Failure, etc)

README.md
```diff
@@ -73,6 +73,25 @@ Whenever the same functionality can be achieved in multiple offerings but in a d
 
 For links within this documentation, please end the link with `/index.md`. This way all links will be checked, and you’ll get an error if a link is broken at any point.
 
+### Concepts
+
+There are several key concepts at Snowplow: events (self-describing, structured), entities, schemas. We must ensure that we use and explain them consistently.
```
jethron, line 78: Nit: "in Snowplow", "at Snowplow" feels directed to staff (in Snowplow the technology vs at Snowplow the company).

docs/understanding-your-pipeline/schemas/index.md
```diff
@@ -1,37 +1,62 @@
...
+To host schemas for your [custom self-described events](/docs/understanding-your-pipeline/events/index.md#self-describing-events) and [custom entities](/docs/understanding-your-pipeline/entities/index.md#custom-entities), you can run your own Iglu, either using [Iglu Server](/docs/pipeline-components-and-applications/iglu/iglu-repositories/iglu-server/index.md) (recommended), or a [static repository](/docs/pipeline-components-and-applications/iglu/iglu-repositories/static-repo/index.md). _(This is not required for Snowplow BDP customers, as Snowplow BDP already includes a private Iglu repository.)_
 
 ## The anatomy of a schema
 
 :::info
 If you are using BDP Cloud, you can create custom schemas using the [Data Structures Builder](/docs/understanding-tracking-design/managing-your-data-structures/builder/index.md) without worrying about how it works under the hood.
 :::
 
-Let’s take a look at an example JSON schema to talk about its constituent parts:
+Snowplow schemas are based on the JSON Schema standard ([draft 4](https://datatracker.ietf.org/doc/html/draft-fge-json-schema-validation-00)). Let’s take a look at an example schema to talk about its constituent parts:
```
jethron, line 59: Maybe also link generally to https://json-schema.org/ as it has some more useful references like the "Learn" section etc.

jethron (commented):
Great changes, love the new reorganisation.

docs/understanding-your-pipeline/entities/index.md
```diff
@@ -0,0 +1,115 @@
...
+:::
+
+Each entity consists of two parts:
+
+- A reference to a [schema](/docs/understanding-your-pipeline/schemas/index.md) that describes the name, version and structure of the entity
+- A set of key-value properties in JSON format — the data associated with the entity
+
+This structure is an example of what we call _self-describing JSON_ — a JSON object with a `schema` and a `data` field.
+
+In the data warehouse, each type of entity gets its own column (or its own table, in case of Redshift). There is no difference between how out-of-the-box and custom entities are stored. See the [structure of Snowplow data](/docs/understanding-your-pipeline/canonical-event/index.md#entities-contexts) for more information.
```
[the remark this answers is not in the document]
stanch: One thing at a time… https://deploy-preview-505--snowplow-docs.netlify.app/docs/storing-querying/schemas-in-warehouse/#versioning ;)

docs/understanding-your-pipeline/events/index.md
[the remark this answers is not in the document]
rlh1994: Yes, I also still feel like you don't need the table at all though.

README.md
```diff
@@ -73,6 +73,25 @@ Whenever the same functionality can be achieved in multiple offerings but in a d
...
+
+**Please, use up-to-date terms:**
+* “Self-describing event”, not “unstructured event”
+* “Entities”, not “contexts”
```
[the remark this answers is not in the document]
stanch: I think it’s just you... I like my smart quotes :P

docs/understanding-your-pipeline/canonical-event/index.md
```diff
@@ -26,9 +27,9 @@ Snowplow data is structured: individual fields are stored in their own columns,
...
 
-For Snowplow users running Amazon Redshift, each custom unstructured event and custom context will be stored in its own dedicated table. These additional tables can be joined back to the core `atomic.events` table, by joining on the `root_id` field in the self-describing event / custom entity table with the `event_id` in the `atomic.events` table, and the `root_tstamp` and `collector_tstamp` field in the respective tables. For users on other warehouses these will be additional columns in the `atomic.events` table.
+For Snowplow users running Amazon Redshift, each self-describing event and entities will be stored in its own dedicated table. These additional tables can be joined back to the core `atomic.events` table, by joining on the `root_id` field in the self-describing event / entity table with the `event_id` in the `atomic.events` table, and the `root_tstamp` and `collector_tstamp` field in the respective tables. For users on other warehouses these will be additional columns in the `atomic.events` table.
```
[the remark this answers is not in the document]
stanch: I actually rewrote all this in #505 but will fix for the time being

README.md
stanch: I do say that in the page about entities:
> especially to refer to a set of _multiple_ entities

However, I am not fond of the idea of making this meaning too official, for two reasons:
* It can create more confusion, because one thing is to move from “context” to “entity” and from “contexts” to “entities” (easy to explain), and another is to move from “context” to “entity” and from “contexts” to “context” (???)
* “Context” is a rather generic and common word, so making it a part of a controlled vocabulary feels hopeless. There will inevitably be sentences like “blah blah this provides some context to the event”. I would prefer these sentences to casually make sense (as they do), rather than mean a very specific thing (especially when this specific thing brings the confusion I mention in the previous point).

Not sure if these are rational arguments but it’s my best attempt to explain why I don’t like this :D
*

rlh1994: I agree we shouldn't really be mixing terms at that level - I think ideally you want the set of lists to be a plural version of list of objects, but then you're left wondering what to call the individual object. We need plural plurals in English is clearly what's causing problems.

You almost want entity<context (list of entities)<contexts (multiple contexts), but I know you don't like that. I think the inverse of that is worse because a single object being a context feels linguistically wrong...

stanch: I’ve been using:
* “entity” to mean a specific entity object
* “entities” to mean a set of entity objects, of the same type or different types
* “types of entities” to mean a set of distinct entity schemas, where it matters to make this distinction (e.g.: “Each type of entity gets its own column”)

Notice that we have the same exact problem with events. We use the word “event” both to mean a particular event and also to mean “type of event” (a “blueprint” for a particular event).

stanch: I am still unsure where “multiple contexts” would be useful beyond the situation where we want to say that event table has multiple columns for each type of entity. (Btw — I just said it and did not need to use the plural of plural!)

rlh1994: But there you are using the singular of entity to describe the list version, which may itself be multiple entity objects! We can't (I mean, we do, but we shouldn't) have the exact same word refer to both an individual instance of a thing (defined by the schema) and the list of that thing which is attached to an event.

rlh1994: But you don't want to. say "...has multiple columns, one for each type of entities" because that's way worse...

README.md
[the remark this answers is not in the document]
jethron: Yep, was just me. Works fine today for some reason. :confounded:

jethron (approved):
LGTM

PR #509, opened 2023-06-28
Add a badge for marking features as new

stanch:


src/css/custom.css
```diff
@@ -284,12 +302,19 @@ article.opensource > a > p::before {
...
+a.menu__link,
+a.card,
+li.menu__list-item,
+li.menu__list-item > div {
   position: relative;
 }
-.enterprise a:hover::after,
-.cloud a:hover::after,
-.opensource a:hover::after {
+li.enterprise > div:hover::after,
```
stanch, line 311: Had to put this `::after` onto a different element, because you can’t have two `::after`s on the same element

mscwilson (approved):
LGTM

The gap between the title and the `new` badge is bigger when the title is long and wraps over multiple lines.
Screenshot 2023-06-28 at 23 20 48

stanch:
> The gap between the title and the `new` badge is bigger when the title is long and wraps over multiple lines.

I tried a few things but looks like we are at the limit of what can be achieved without swizzling the sidebar item (which I’d rather not do as it creates a lot of code). In any case, I guess we will not have so many things marked as new in a row, nor should we have such long sidebar labels.

PR #516, opened 2023-07-04
Interactive dbt configs

rlh1994:
Add the remaining schemas for interactive config generator for dbt packages


src/components/JsonSchemaValidator/dbtFractribution.js
```diff
@@ -0,0 +1,119 @@
...
+            "Remove all occurances of the channel, if it isn't the only channel",
+        },
+        unique_path: {
+          type: 'string',
+          title: 'Unique Path',
+          description: 'No change',
+          enum: ['null'],
+        },
+      },
+    },
```
georgewoodhead, line 85: enum of only null on most of these doesn't seem right?

src/components/JsonSchemaValidator/dbtFractribution.js
```diff
@@ -0,0 +1,119 @@
...
+      title: 'Consider Intra-sessions Channels',
+    },
+    snowplow__run_python_script_in_snowpark: {
+      type: 'boolean',
+      title: '(Snowflake) Run the python script via Snowpark',
+    },
+    snowplow__attribution_model_for_snowpark: {
+      type: 'string',
+      title: '(Snowflake) Attribution Model type',
+    },
```
georgewoodhead, line 117: Maybe add enum here for `shapley`, `first_touch` etc.

georgewoodhead (changes requested):
Nice stuff! Left a few small comments

src/components/JsonSchemaValidator/dbtFractribution.js
rlh1994: You'd think so, but the thing is most of them don't actually have any parameters - however if I don't give them the option to have _some_ value I can't get them to render in the output. An empty string is also wrong, so this was the best way I could find to get it to work...

rlh1994:
@zloff Can you validate the fract stuff please as I just took a best attempt at it

src/components/JsonSchemaValidator/dbtFractribution.js
```diff
@@ -0,0 +1,120 @@
...
+      minimum: 1,
+    },
+    snowplow__path_lookback_steps: {
+      type: 'integer',
+      title: 'Path Lookback Steps',
+      description:
+        'Maximum number of steps before the conversion (0 is unlimited)',
+      minimum: 0,
+    },
+    snowplow__path_transforms: {
```
zloff, line 45: I know how the frac package works with the path transforms and I still got confused by this section. i'm not sure how difficult it is to add some instruction under the snowplow__path_transforms heading, but I think we need to give some more instruction. I understand why, but it's especially confusing having some be dropdowns with null as the option and some text entry without much info as to why unless you go trawl through the udfs.sql file

src/components/JsonSchemaValidator/dbtFractribution.js
```diff
@@ -0,0 +1,120 @@
...
+        exposure_path: {
+          type: 'string',
+          title: 'Exposure path',
+          description: 'Consequitive same channels are reduced to one',
+          enum: ['null'],
+        },
+        first_path: {
+          type: 'string',
+          title: 'First Path',
+          description: 'First occurance of path is kept',
```
zloff, line 57: sorry, occurrence

zloff (commented):
Looking good, I really like the config builder

docs/modeling-your-data/modeling-your-data-with-dbt/dbt-configuration/ecommerce/index.md
```diff
@@ -12,7 +12,6 @@ import TabItem from '@theme/TabItem';
 
 This package utilizes a set of variables that are configured to recommended values for optimal performance of the models. Depending on your use case, you might want to override these values by adding to your `dbt_project.yml` file.
 
-
 :::note
 
 All variables in Snowplow packages start with `snowplow__` but we have removed these in the below table for brevity.
```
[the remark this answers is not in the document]
rlh1994: The main reason for this is mostly that the table ends up really wide and you have to scroll horizontally to see the useful information then. We've not had any issues with anyone asking about an issue and this being the problem, plus if they look at any of the variables in the package they will see they all start with that. I've swapped it out to a caution though.

src/components/JsonSchemaValidator/dbtFractribution.js
rlh1994: I did however not correctly understand how types and values work in jsonschema, which is now fixed

georgewoodhead (approved):
Bill Murray LGTM

docs/modeling-your-data/modeling-your-data-with-dbt/dbt-configuration/ecommerce/index.md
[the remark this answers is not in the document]
zloff: Ah ok, sound good

zloff (approved):

PR #494, opened 2023-06-22
Revamp AWS quick-start documentation to include Redshift + Databricks

jbeemster:
@stanch this is a big PR but it updates our AWS quick-start to be in-line with the latest modules and I think greatly simplifies a few of the instructions.

It also crucially now supports Redshift + Databricks support with the Terraform on the back being able to easily extend to new targets as needed.

@stanch I could not find the source for the images used in the "/docs/getting-started-on-snowplow-open-source/quick-start-aws/summary-of-what-you-have-deployed/?warehouse=postgres" page so have copied the Snowflake one.  It will need a minor tweak for both of the new destinations.

Related PR: https://github.com/snowplow/quickstart-examples/pull/60

---

Deprecation PR: https://github.com/snowplow-devops/terraform-aws-snowflake-loader-setup/pull/7

PR #529, opened 2023-07-10
Remove manual setup docs

stanch:
These docs are quite outdated, and the quick start will soon cover almost all relevant setups (see #494).


Andy-Hay (approved):
LGTM - i went through the preview and it looks to make more sense.  There was a lot of repetition in the manual setup pages which is nicely removed, keeping us with fewer (hopefully one) sources of truth.

PR #545, opened 2023-07-19
docs: remove rdbms command prefix for table-check

pkutaj:
* Not sure about the wider context, but the current version of igluctl (0.11.x) does not support `rdbms` command. The `table-check` command can be called only without that prefix. 

```
./igluctl --help
Usage:
    igluctl static
    igluctl lint
    igluctl server
    igluctl --version
    igluctl table-check
```

* @pondzix could you please confirm?


stanch (approved):
Confirming, thanks

PR #554, opened 2023-07-27
Fix broken link on the Adding and removing users page

andreamarcolin:


stanch (approved):
Apparently broken since the Wordpress days!

PR #561, opened 2023-08-02
Enable tracking scenarios for cloud customers

gleb-lobov:
For the data structure change - I've added a link to the Cloud specific documentation, because main one was for the 
Enterprise

### DIFF
--- a/docs/feature-comparison/index.md
+++ b/docs/feature-comparison/index.md
@@ -47,7 +47,7 @@ To find out more about the support services offered to Snowplow BDP customers se
 | <h3>Collaborate across multiple teams</h3> | [Open Source](/docs/getting-started-on-snowplow-open-source/index.md) | [BDP Cloud :new:](/docs/getting-started-on-snowplow-bdp-cloud/index.md) | [BDP Enterprise](/docs/getting-started-on-snowplow-bdp-enterprise/index.md) |
 | Seats | N/A | unlimited | unlimited |
 | [Data structures tooling & API](/docs/understanding-tracking-design/managing-your-data-structures/ui/index.md) | ❌ | ✅<br/>_(UI only)_ | ✅ |
-| [Tracking scenarios](/docs/understanding-tracking-design/tracking-plans/index.md) | ❌ | ✅ | ✅ |
+| [Tracking scenarios](/docs/understanding-tracking-design/tracking-plans/index.md) | ❌ | ✅<br/>_(UI only)_ | ✅ |
 | [Data modeling management tooling](/docs/modeling-your-data/running-data-models-via-snowplow-bdp/dbt/using-dbt/index.md) | ❌ | _coming soon_ | ✅ |
 | [Tracking catalog](/docs/discovering-data/tracking-catalog/index.md) | ❌ | ❌ | ✅ |
 | <h3>Deployment & security</h3> | [Open Source](/docs/getting-started-on-snowplow-open-source/index.md) | [BDP Cloud :new:](/docs/getting-started-on-snowplow-bdp-cloud/index.md) | [BDP Enterprise](/docs/getting-started-on-snowplow-bdp-enterprise/index.md) | 
--- a/docs/feature-comparison/index.md
+++ b/docs/feature-comparison/index.md
@@ -46,7 +46,7 @@ To find out more about the support services offered to Snowplow BDP customers se
 | Outage protection | ❌ | ❌ | ✅<br/>_(top tier only)_ |
 | <h3>Collaborate across multiple teams</h3> | [Open Source](/docs/getting-started-on-snowplow-open-source/index.md) | [BDP Cloud :new:](/docs/getting-started-on-snowplow-bdp-cloud/index.md) | [BDP Enterprise](/docs/getting-started-on-snowplow-bdp-enterprise/index.md) |
 | Seats | N/A | unlimited | unlimited |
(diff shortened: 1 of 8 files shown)


docs/feature-comparison/index.md
```diff
@@ -46,8 +46,8 @@ To find out more about the support services offered to Snowplow BDP customers se
 | Outage protection | ❌ | ❌ | ✅<br/>_(top tier only)_ |
 | <h3>Collaborate across multiple teams</h3> | [Open Source](/docs/getting-started-on-snowplow-open-source/index.md) | [BDP Cloud :new:](/docs/getting-started-on-snowplow-bdp-cloud/index.md) | [BDP Enterprise](/docs/getting-started-on-snowplow-bdp-enterprise/index.md) |
 | Seats | N/A | unlimited | unlimited |
-| [Data structures tooling & API](/docs/understanding-tracking-design/managing-your-data-structures/ui/index.md) | ❌ | ✅<br/>_(UI only)_ | ✅ |
-| [Tracking scenarios](/docs/understanding-tracking-design/tracking-plans/index.md) | ❌ | ❌ | ✅ |
+| [Data structures tooling & API](/docs/understanding-tracking-design/managing-your-data-structures/ui/index.md) | ❌ | ✅<br/>[_(UI only)_](/docs/understanding-tracking-design/managing-your-data-structures/builder/) | ✅ |
```
stanch, line 49:
```suggestion
| [Data structures tooling & API](/docs/understanding-tracking-design/managing-your-data-structures/ui/index.md) | ❌ | ✅<br/>[_(UI only)_](/docs/understanding-tracking-design/managing-your-data-structures/builder/index.md) | ✅ |
```
See README.md for this repo :)

gleb-lobov: Whoops 😅 Thanks for noticing!

docs/understanding-tracking-design/managing-your-tracking-scenarios/index.md
```diff
@@ -5,6 +5,7 @@ sidebar_position: 95
 sidebar_custom_props:
   offerings:
     - enterprise
+    - cloud
```
stanch, line 8: Wouldn’t this mislead people into thinking the API is available for Cloud? Probably we need to move the offering definition to the child pages, where one would be enterprise + cloud and the other just enterprise. (The way I implemented this mechanism, it does not currently support defining something and then refining in a child page, although it’s of course possible to change.)

gleb-lobov: Good catch again. I've altered the implementation, to be able to override the parent offerings. Please check if the implementation makes sense, my JS skills are straight from the first page of Google.

One side effect that I've noticed - is that child pages now got enterprise/cloud emoji assigned.

Old:
image
New:
image

Let me know if anything needs to be done about the implementation/emojies

gleb-lobov: @stanch Elisa says we are ready to merge this change, could you take a look?

docs/understanding-tracking-design/managing-your-tracking-scenarios/api/index.md
```diff
@@ -2,6 +2,10 @@
 title: "Managing tracking scenarios via the API"
 sidebar_label: "Using the API 🆕"
 sidebar_position: 2
+sidebar_custom_props:
+  offerings:
+    - enterprise
+    - cloud
```
stanch, line 8: Is the API available on Cloud?

docs/understanding-tracking-design/managing-your-tracking-scenarios/ui/index.md
```diff
@@ -2,6 +2,9 @@
 title: "Managing tracking scenarios in the Console"
 sidebar_label: "Using the UI 🆕"
 sidebar_position: 1
+sidebar_custom_props:
+  offerings:
+    - enterprise
```
stanch, line 7: I thought it was the opposite — UI on enterprise and cloud, and API enterprise only?

src/theme/MDXContent/index.js
```diff
@@ -24,7 +24,12 @@ export default function MDXContentWrapper(props) {
 
   const legacy = _.some(_.initial(breadcrumbs), item => item.customProps?.legacy)
   const outdated = !legacy && _.some(_.initial(breadcrumbs), item => item.customProps?.outdated)
-  const offerings = _.find(breadcrumbs, item => item.customProps?.offerings)
+  let offerings; 
+  if (breadcrumbs) {
+    const currentPage = breadcrumbs.slice(-1)
+    const parents = breadcrumbs.slice(0, -1)
+    offerings = _.find(currentPage.map(x => Array.isArray(x.customProps?.offerings)).includes(true) ? currentPage : parents, item => item.customProps?.offerings)
+  }
```
stanch, line 32:
```suggestion
 const offerings = _.findLast(breadcrumbs, item => item.customProps?.offerings)
```

I think this might be a bit simpler ;) Also your version seems to only defer to the immediate parent? Would it work when a page overriding the `offerings` is nested deeper?

stanch (changes requested):
Two things:

First, the tooltips are not showing correctly (see screenshot). I think we might be able to fix this by playing with the css a bit. I’ll give it a try.
image

Second, could you add `text-wrap: nowrap;` to the `::before` class to prevent it from showing the two icons vertically?

stanch (approved):

PR #562, opened 2023-08-02
Add docs for retry capability on the PHP tracker

matus-tomlein:
Adds documentation for the new retry logic in the socket and sync emitter on the PHP tracker.


greg-el (approved):
LGTM

PR #570, opened 2023-08-09
Update media player schema and versions of mobile trackers

matus-tomlein:
This PR updates the schema URI for the media player entity. Also update versions for the mobile trackers.


mscwilson (approved):
lgtm

PR #571, opened 2023-08-09
Remove gendered language and americanize spellings

rlh1994:
Removes the use of gendered language anywhere I could find it, and fixed any other spelling issues on the pages I found.


docs/collecting-data/collecting-from-own-applications/flutter-tracker/initialization-and-configuration/index.md
```diff
@@ -55,15 +54,15 @@ The `EmitterConfiguration` class was added in v0.3.0.
...
+Activity tracking monitors whether a user continues to engage with a page over time, and record how they digests content on the page over time. That is accomplished using 'page ping' events. If activity tracking is enabled, the web page is monitored to see if a user is engaging with it. (E.g. is the tab in focus, does the mouse move over the page, does the user scroll etc.) If any of these things occur in a set period of time (`minimumVisitLength` seconds from page load and every `heartbeatDelay` seconds after that), a page ping event fires, and records the maximum scroll left / right and up / down in the last ping period. If there is no activity in the page (e.g. because the user is on a different tab in their browser), no page ping fires.
```
mscwilson, line 65:
```suggestion
Activity tracking monitors whether a user continues to engage with a page over time, and record how they digest content on the page over time. That is accomplished using 'page ping' events. If activity tracking is enabled, the web page is monitored to see if a user is engaging with it. (E.g. is the tab in focus, does the mouse move over the page, does the user scroll etc.) If any of these things occur in a set period of time (`minimumVisitLength` seconds from page load and every `heartbeatDelay` seconds after that), a page ping event fires, and records the maximum scroll left / right and up / down in the last ping period. If there is no activity in the page (e.g. because the user is on a different tab in their browser), no page ping fires.
```

mscwilson (approved):
lgtm

docs/collecting-data/collecting-from-own-applications/javascript-trackers/browser-tracker/browser-tracker-v3-reference/tracking-events/index.md
```diff
@@ -121,9 +120,9 @@ The page view and every subsequent page ping will have both a static_context and
 
 ### Activity Tracking: page pings
 
-As well as tracking page views, we can monitor whether a user continues to engage with a page over time, and record how he / she digests content on the page over time.
+As well as tracking page views, we can monitor whether a user continues to engage with a page over time, and record how they digest content on the page over time.
```
stanch, line 123:
```suggestion
As well as tracking page views, we can monitor whether users continue to engage with pages over time, and record how they digest content on each page over time.
```

src/theme/SearchBar/index.js
```diff
@@ -55,7 +55,7 @@ function DocSearch({ contextualSearch, externalUrlRegex, ...props }) {
       mergeFacetFilters(contextualSearchFacetFilters, configFacetFilters)
     : // ... or use config facetFilters
       configFacetFilters
-  // We let user override default searchParameters if she wants to
+  // We let user override default searchParameters if they want to
```
stanch, line 58:
```suggestion
  // We let users override default searchParameters if they want to
```

stanch (commented):
Nothing against singular “they”, but in the cases where plural can be used, I think it reads a bit nicer.

I only made each suggestion once but obviously they apply to other instances of the same text.

PR #586, opened 2023-08-21
Tweaks for new GCP deployment

jbeemster:
https://github.com/snowplow/quickstart-examples/pull/68

@stanch this pulls all GCP modules up to date with how we deploy Azure + LICENSE changes across all of them.  These PRs need to be merged concurrently.

### DIFF
--- a/docs/getting-started-on-snowplow-open-source/quick-start/index.md
+++ b/docs/getting-started-on-snowplow-open-source/quick-start/index.md
@@ -585,7 +585,7 @@ terraform plan
 terraform apply
 ```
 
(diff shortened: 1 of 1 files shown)


jbeemster:
> The page mentions postgres_db_address, but here you have postgres_db_ip_address. Are they meant to be different on AWS and GCP (e.g. because the AWS one is a domain name?) If so we should mention both.

Indeed on GCP its an IP address instead of a domain name.  Both should be mentioned yes!

> The BigQuery database used to be <prefix>_snowplow_db and is now <prefix>_pipeline_db. I kinda prefer the old name... But in any case needs changing either here or in Terraform. It is also mentioned [on this page](https://docs.snowplow.io/docs/getting-started-on-snowplow-open-source/what-is-deployed/?warehouse=bigquery#loaders).

In fact it was already `pipeline_db` in the `default` pipeline deployment (secure had `snowplow_db` so normalized it).  Happy to revert back to `snowplow_db` though.

stanch (approved):

PR #594, opened 2023-08-29
fix: Redshift file format TSV > JSON

pkutaj:
I believe this is a typo, could you please check and merge it? Thanks!


stanch:
Hi @pkutaj, the diagram is correct as stated.

PR #598, opened 2023-09-04
Bump version for the ecommerce and media dbt package versions

matus-tomlein:
Updates the component versions for media player v0.5.3 and ecommerce 0.5.2.


rlh1994 (approved):
A complicated one, but I just about understand it!

PR #607, opened 2023-09-15
Add Synapse Analytics to the user guide

stanch:
Still missing the connection instructions at https://docs.snowplow.io/docs/first-steps/querying/, but that will likely be a separate PR, together with a Quick Start update.


docs/storing-querying/storage-options/index.md
```diff
@@ -40,12 +46,41 @@ Data warehouses and data lakes are primary destinations for Snowplow data. For o
...
+| Databricks | Micro-batching<br/>_(via a [data lake](#data-lake-loaders))_ | LAKELOADERLINK | Early release |
+| Synapse Analytics | Micro-batching<br/>_(via a [data lake](#data-lake-loaders))_ | LAKELOADERLINK | Early release |
 
   </TabItem>
 </Tabs>
 
-### Data lake loaders
+## Data lake loaders
+
+All lake loaders are micro-batching.
```
stanch, line 59: Actually is that true for S3/GCS loaders?

docs/storing-querying/querying-data/index.md
```diff
@@ -265,7 +308,7 @@ select
     ...
 from
     unique_events
-where 
+where
```
istreeter, line 311: We could captialize these too.  Mentioning it because you've already done some capitalization fixes elsewhere in this PR.

docs/storing-querying/storage-options/index.md
```diff
@@ -40,12 +46,41 @@ Data warehouses and data lakes are primary destinations for Snowplow data. For o
...
+
+<Tabs groupId="cloud" queryString>
+  <TabItem value="aws" label="AWS" default>
+
+| Lake | Format | Compatibility | Loader application | Status |
+| --- | --- | --- | --- | --- |
+| S3 | TSV | Amazon Athena | [S3 Loader](/docs/pipeline-components-and-applications/loaders-storage-targets/s3-loader/index.md) | Production-ready |
```
istreeter, line 66: IMO we should not be encouraging users to query TSV with Athena.  It's a not production-appropriate way to query data.

At this point, the only reason to use the S3 loader is as a pre-cursor to the RDB batch transformer.  I do not see any other reason to use it.

**correction** -- actually, we should advise it a solution for querying failed events (because we don't have any better solution).  But not enriched events.

docs/storing-querying/storage-options/index.md
```diff
@@ -40,12 +46,41 @@ Data warehouses and data lakes are primary destinations for Snowplow data. For o
...
+
+  </TabItem>
+  <TabItem value="gcp" label="GCP">
+
+| Lake | Format | Compatibility | Loader application | Status |
+| --- | --- | --- | --- | --- |
+| GCS | Delta | Databricks | LAKELOADERLINK | Early release |
+| GCS | TSV | BigQuery | [GCS Loader](/docs/pipeline-components-and-applications/loaders-storage-targets/google-cloud-storage-loader/index.md) | Only recommended for [failed events](/docs/understanding-your-pipeline/failed-events/index.md) |
```
istreeter, line 74: Failed events are JSON format, not TSV.

The GCS loader is similar to the S3 loader, in that it doesn't care about the format.  But if we're recommending it only for failed events, then JSON is the only sensible thing to put in the second column.

docs/storing-querying/querying-data/index.md
```diff
@@ -98,30 +94,46 @@ Each type of self-describing event is in a dedicated `OBJECT`-type column. The c
...
+You can query fields in the self-describing event like so:
+
+```sql
+SELECT
+    ...
+    JSON_VALUE(unstruct_event_my_example_event_1, '$.myField')
+    ...
+FROM
+    OPENROWSET(...) AS [result]
+```
```
rlh1994, line 135: I might have missed it, but I can't find anywhere we actually give the user an example of how to use `OPENROWSET` or even at least link to the microsoft docs?

docs/storing-querying/querying-data/index.md
```diff
@@ -186,52 +201,80 @@ Each type of entity is in a dedicated `ARRAY`-type column. The column name will
...
+You can query a single entity’s fields by extracting them like so:
+
+```sql
+SELECT
+    ...
+    JSON_VALUE(contexts_my_entity_1, '$[0].myField')
+    ...
+FROM
+    OPENROWSET(...) AS [result]
+```
```
rlh1994, line 266: I can't actually test this and haven't used T-SQL before, so I assume it's working.

docs/storing-querying/querying-data/index.md
```diff
@@ -376,5 +419,27 @@ from atomic.events a
...
+        OPENROWSET(...) as [result]
+)
+
+select
+    ...
+from
+    unique_events
+where
+    event_id_dedupe_index = 1
+```
```
rlh1994, line 442: These ones are all lowercase still

rlh1994 (approved):
A few nitpicks

docs/storing-querying/querying-data/index.md
stanch: That will be in the connection instructions on a different page... After all I am probably adding that in this review

docs/storing-querying/querying-data/index.md
stanch: I have tested it

docs/storing-querying/querying-data/index.md
stanch: your fault!

rlh1994: Contributions are welcome

docs/storing-querying/querying-data/index.md
stanch: @rlh1994 I expanded the queries here, and the details on where to get `<events>` are in #611.

istreeter (approved):

docs/storing-querying/loading-process/index.md
```diff
@@ -29,18 +30,32 @@ We load data into BigQuery using the [BigQuery Loader](/docs/pipeline-components
 <BigQueryLoaderDiagram/>
 
   </TabItem>
-  <TabItem value="databricks" label="Databricks">
+  <TabItem value="databricks" label="Databricks (direct)">
```
istreeter, line 33: I understand your word choice, but in a way this loading method is less direct.

PR #611, opened 2023-09-18
Add Lake Loader to Quick Start

stanch:


docs/getting-started-on-snowplow-open-source/quick-start/index.md
```diff
@@ -413,7 +421,7 @@ CREATE SCHEMA IF NOT EXISTS ${schema_name}
 
 The security principal used by the loader needs a `Databricks SQL access` permission, which can be enabled in the _Admin Console_.
 
-Databricks does not have table access enabled by default. Enable it with an 
+Databricks does not have table access enabled by default. Enable it with an
```
istreeter, line 424: I know it's not the main point of your PR.  But.... you removed a space from a nearby sentence, so here goes.....

This section refers to a Databricks feature [which is described on their docs as "legacy"](https://docs.databricks.com/en/data-governance/table-acls/table-acl.html).  I wonder if we can drop this initialization script from our docs.  And say if users are on a legacy setup then they just need to figure this out for themselves.

docs/getting-started-on-snowplow-open-source/quick-start/index.md
```diff
@@ -615,4 +628,108 @@ For solutions to some common Terraform errors that you might encounter when runn
 
 :::
 
+## Configure the destination
+
+<Tabs groupId="warehouse" queryString>
+  <TabItem value="postgres" label="Postgres" default>
```
istreeter, line 634: Also not the main point of your PR but....

To me it looks out-of-place that Postgres is part of the quick start.  It made sense 2 years ago, because we had no other viable streaming loader.  But now we have multiple other loaders in the quick start, I don't think this should even be part of the product that we advertise.

istreeter (approved):

docs/getting-started-on-snowplow-open-source/quick-start/index.md
istreeter: I think so yes.  But ideally we should test out this section end-to-end, with that part excluded.  Nobody currently still at Snowplow has ever run these steps.

Sorry to hijack your PR which is about something different!  We shouldn't let this be a blocker to merging the current PR.  But I also don't want to forget about it entirely.

docs/getting-started-on-snowplow-open-source/quick-start/index.md
stanch: Discussed on Slack — keeping Postgres for the time being

docs/understanding-your-pipeline/destinations/index.md
```diff
@@ -15,7 +15,7 @@ Here you will find an overview of what’s available.
 
 ## Data warehouses and lakes
 
-Snowplow is primarily built for data warehouse and lake destinations and supports Redshift, BigQuery, Snowflake and Databricks, as well as S3 and GCS, via the [various loaders](/docs/storing-querying/storage-options/index.md).
+Snowplow is primarily built for data warehouse and lake destinations and supports Redshift, BigQuery, Snowflake, Databricks and Synapse Analytics, as well as S3, GCS and ADLS, via the [various loaders](/docs/storing-querying/storage-options/index.md).
```
jbeemster, line 18: Are we missing Azure Fabric from this list or is that inferred from ADLS?

docs/understanding-your-pipeline/architecture-overview/index.md
```diff
@@ -15,7 +15,7 @@ The diagram above illustrates a typical Snowplow pipeline with data flowing left
...
 - Once validated, each event is enriched by the [**Enrichments**](/docs/enriching-your-data/available-enrichments/index.md) you have configured for your pipeline.
 - The enriched events are sent to real-time streams (Kinesis on AWS, PubSub on GCP). At this point, data can be [**forwarded**](/docs/destinations/forwarding-events/index.md) to other applications to power real-time use cases.
-- [**Loaders**](/docs/storing-querying/storage-options/index.md) then load your data off the real-time streams into the various destinations you have set up for your pipeline. Typically these include a data warehouse (Redshift, Snowflake, BigQuery, Databricks) but can also include other destinations.
+- [**Loaders**](/docs/storing-querying/storage-options/index.md) then load your data off the real-time streams into the various destinations you have set up for your pipeline. Typically these include a data warehouse (Redshift, Snowflake, BigQuery, Databricks) or a data lake (S3, GCS, ADLS). Note that you can also connect a data warehouse to a data lake (e.g. Synapse Analytics + ADLS) instead of loading to a warehouse directly.
```
jbeemster, line 18: Similarly here should we call our Azure Fabric?

docs/understanding-your-pipeline/destinations/index.md
stanch: Added OneLake — with Fabric things get a bit muddy because it’s not really a warehouse. But since OneLake is a part of Fabric, it will imply Fabric.

PR #612, opened 2023-09-20
Snowbridge default config path

colmsnowplow:
Snowbridge 2.2.0 has a default config path in the docker image - amend docs to suit.


colmsnowplow:
Nope, that describes running CLI, whereas the config default path is added in docker image.

stanch (approved):

PR #630, opened 2023-10-04
add dbt utils sql options

rlh1994:


agnessnowplow:
Not for this PR but I was thinking we might use all the title line space to educate the user at the same time, getting them used to what the variables are called by adding the actual variable name as well in bracket or if no title, just add them there? 

Also, I would highlight that the schema generation will be a gradual process and they will see it being formed below (the code expands so much, for first time users it can be really confusing where and how it is generated. Plus if they don't see anything changed it means that the value they are given are the defaults, therefore they don't have to add anything to their project.yml related to that.

I can add something like that in the unified PR btw as this impacts all and I want to avoid merge conflicts as much as I can.

image

agnessnowplow (approved):

PR #646, opened 2023-10-16
Update form tracking docs for JavaScript trackers v3.16.0

adatzer:


docs/collecting-data/collecting-from-own-applications/javascript-trackers/browser-tracker/browser-tracker-v3-reference/tracking-events/index.md
```diff
@@ -964,12 +964,58 @@ This is a function used to determine which elements are tracked. The element is
 
 This is a function used to transform data in each form field. The value and element (2.15.0+ only) are passed as arguments to the function and the tracked value is replaced by the value returned.
 
+**In versions prior to 3.16.0**, the transform function would receive 2 arguments, that were different between `submit_form` and `change_form` or `focus_form` events. More specifically, the transform function signature **was**:
```
matus-tomlein, line 967: Can we put this prior to 3.16 section to a `:::note` admonition? Just to visually separate it from the current behaviour.

So something like
```
:::note Behaviour prior to version 3.16 of the tracker
...
:::
```

docs/collecting-data/collecting-from-own-applications/javascript-trackers/browser-tracker/browser-tracker-v3-reference/tracking-events/index.md
```diff
@@ -964,12 +964,58 @@ This is a function used to determine which elements are tracked. The element is
...
+```typescript
+interface ElementData extends Record<string, string | null | undefined> {
+  name: string;
+  value: string | null;
+  nodeName: string;
+  type?: string;
+}
+```
+
+**Since version 3.16.0**, in order to allow the transform function to consistently access all attributes of the tracked element in a backwards compatible way, the transform function receives an extra argument which is the tracked HTML element in all form tracking events. The transform function signature is now:
```
matus-tomlein, line 984: Nit: a minor thing, but this is not a changelog so we shouldn't argue for why something was changed or not, users don't care. I think we should just explain how things work without too much talk around.

```suggestion
**Since version 3.16.0**, the transform function receives three arguments:
1. The value of the element.
2. Either the HTML element (for `change_form` and `focus_form` events) or an instance of `ElementData` (for `submit_form` events).
3. The HTML element (in all form tracking events).
The function signature is:
```

matus-tomlein (approved):
LGTM! Just a couple of nitpicky comments about the style of the explanation.

igneel64 (approved):

adatzer:
Thanks for reviewing @matus-tomlein and @igneel64 ! Made the recommended changes and also rebased onto main.

PR #650, opened 2023-10-18
Update Snowplow tracker protocol to reflect current tracking including OOTB self-describing events, entities and link to modeling

matus-tomlein:
This PR updates the [Snowplow Tracker Protocol page](https://deploy-preview-650--snowplow-docs.netlify.app/docs/collecting-data/collecting-from-own-applications/snowplow-tracker-protocol/) with a more up-to-date reflection of the tracked event data:

1. tries to make it easier to consume – adds structure, each topic has a separate page,
2. updates the information to include new e-commerce, mobile, media data and also document self-describing events and context entities,
3. adds a reusable component (`<SchemaProperties />`) for describing events and context entities using their schemas – this shows their URI, description, gives an example payload and also shows an SQL query for querying events in warehouses (still missing the query for context entities).
4.  moves the focus away from building a custom tracker but more about understanding the Snowplow events and entities.

@agnessnowplow has extended the [Canonical event model page](https://deploy-preview-650--snowplow-docs.netlify.app/docs/understanding-your-pipeline/canonical-event/#out-of-the-box-fields) to add columns for whether each property is tracked on Web/mobile and add links to the new pages under the Snowplow events section


docs/collecting-data/collecting-from-own-applications/snowplow-tracker-protocol/app-information/index.md
```diff
@@ -0,0 +1,81 @@
+---
+title: "App & tracker information"
+---
+
+Information about the app and the tracker instance that the events originate from is tracked using:
+
+1. Atomic event properties
```
agnessnowplow, line 7: I feel like we are missing an explanation of the out of the box atomic event properties, they appear hidden in fragments like this, sometimes clearly categorised under atomic events like here, sometime under a category called "platform" (for browser related fields) I feel like it would help if we gave an intro and explained what group of fields we are tracking for web and mobile out of the box and which not. If we leave that to the other page I started updating then at least we should link that to the main tracker page and explain our documentation structure at the beginning.

docs/collecting-data/collecting-from-own-applications/snowplow-tracker-protocol/user-and-session-identification/index.md
```diff
@@ -0,0 +1,29 @@
+---
+title: "User and session identification and stitching"
```
agnessnowplow, line 2: if it is called stitching we should all out that they should track user_ids because that will guarantee out of the box user stitching for modelling. otherwise i would remove it from the name

agnessnowplow (commented):
I only had a quick look but so far I'm loving the structure changes you made! I love the fact that we have the dedicated pages for each context they deserved and also the main page where you grouped them by theme (e.g. device related). One thing I'm still thinking about is we have a long list of pages under Trackers/Snowplow events. It would be nice to group them somehow. Also feel like we do not really highlight for which of them they get out of the box modelling which could be something for them to consider.

docs/collecting-data/collecting-from-own-applications/snowplow-tracker-protocol/self-describing-events/index.md
```diff
@@ -0,0 +1,68 @@
...
+
+Historically, custom self-describing events were called “unstructured” and the legacy custom events were called “structured”. This terminology can be confusing, so we don’t use it anymore. However, you might find its remnants in some of the APIs.
+
+:::
+
+An example of a self-describing event for a product view event:
+
+```json
+{
+  "schema": "iglu:com.my_company/viewed_product/jsonschema/1-0-0",
```
agnessnowplow, line 19: Wouldn't this need to contain `self-desc` there? I have no idea how it works but I had trouble understanding the diff based on this example alone and I saw that in one of the implementations. (cwv i think)

mscwilson:
This is a great idea.
I haven't looked thoroughly through the pages since maybe they're under construction? I agree with Agnes that some more structure is needed, it's a long list of pages currently. Can they be in sections to match the sections of the top-level page? Custom events/entities, OOTB events, info with events.

docs/collecting-data/collecting-from-own-applications/snowplow-tracker-protocol/app-error-events/index.md
```diff
@@ -0,0 +1,125 @@
...
+
+Exception (error) tracking captures any unhandled exceptions within the application.
+
+ | 
+---|---
+Type | Event
+Schema | `iglu://com.snowplowanalytics.snowplow/application_error/jsonschema/1-0-2` 
+Web | ✅
+Mobile | ✅
+Atomic table field name | `unstruct_event_com_snowplowanalytics_snowplow_application_error_1`
```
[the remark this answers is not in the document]
matus-tomlein: I have added a "How to query?" section which gives a query for each warehouse. Currently it only works for events, need to add support for context entities as well.

docs/collecting-data/collecting-from-own-applications/snowplow-tracker-protocol/ootb-data/user-and-session-identification/index.md
```diff
@@ -0,0 +1,177 @@
...
+
+In Snowplow events, sessions are tracked on the client-side – by the tracker.
+The session information is then attached to the events when they are tracked.
+
+The session information consists of two main pieces of information (when using the client session entity discussed below, there is also extra information tracked with sessions):
+
+1. **Session identifier**. UUID identifier generated by the tracker at the start of the session.
+2. **Session index**. Index of the session for the current user (per the tracker generated user identifier).
+
+Sessions have a configurable inactivity timeout.
```
agnessnowplow, line 77: Is there a default / suggested default to mention here?

docs/collecting-data/collecting-from-own-applications/snowplow-tracker-protocol/going-deeper/example-requests/index.md
```diff
@@ -1,7 +1,5 @@
 ---
 title: "Example Tracker Requests"
-date: "2022-10-10"
-sidebar_position: 100
 ---
 
 # Example Tracker Requests
```
miike, line 5: I'd do an infobox here just describing that the trackers will perform these network requests for you but if you want more detail, or to write a new tracker implementation then this page is helpful.

miike (approved):

docs/collecting-data/collecting-from-own-applications/snowplow-tracker-protocol/ootb-data/user-and-session-identification/index.md
matus-tomlein: Good point, have mentioned the 30 mins default.

PR #654, opened 2023-10-24
Feature/versioned dbt schema

rlh1994:
Draft PR for now, still quite a bit to do to get this to work:
- [x] Ideally split the components but keep the state method so that we can interweave them with markdown
- [x] finish copying over definitions and defaults, work out what to do with links
- [x] add the MUI key
- [x] add some historic schemas?
- [x] add some comments because there is 0 chance I will understand this in a few months

The goal of this PR is to support versioning of our dbt configuration schemas, this will be useful internally for upcoming console features, as well as to all customers who are running older versions (and us supporting them) without having to check the package yaml file.

The expectation is that the only things that will need to be edited directly on the config pages going forward is:
1. The versions that we have schemas for
2. The schema setter part

Everything else will be managed directly via the schemas, including the massive table, the group that variables belong to, and the config generator. In theory the tables are nicer (searchable!), and we can more easily add more thinks as needed in the future via fields in the schemas and changes to the react functions.


docs/modeling-your-data/modeling-your-data-with-dbt/dbt-configuration/normalize/index.mdx
```diff
@@ -0,0 +1,111 @@
...
+        <h2 id="Variables">Package Configuration Variables</h2>
+        <p>This package utilizes a set of variables that are configured to recommended values for optimal performance of the models. Depending on your use case, you might want to override these values by adding to your <code>dbt_project.yml</code> file.</p>
+        <Admonition type="note">
+          All variables in Snowplow packages start with `snowplow__` but we have removed these in the below tables for brevity.
+        </Admonition>
+        <br/>
+        {dbtConfigOutput[2]} {/* JsonToTable */}
+        <br/>
+        <Admonition type="tip">
+          When modifying the <code>session/user_identifiers</code> or using <code>session/user_sql</code> in the normalize package these will overwrite the <code>domain_sessionid</code> and <code>domain_userid</code> fields in tables, rather than being <code>session/user_identifier</code> as in the core utils implementation. This is for historic reasons to mitigate breaking changes. Original values for these fields can be found in `original_domain_session/userid` in each table.
```
rlh1994, line 69: This stuff doesn't exist in normalize (or most packages other than web for this version actually) so we should remove it until the base package is added in.

emielver (approved):
LGTM!

docs/modeling-your-data/modeling-your-data-with-dbt/dbt-models/dbt-unified-data-model/conversions/index.md
```diff
@@ -98,7 +69,7 @@ The `snowplow__conversion_events` variable in our project takes a list of dictio
 
 :::tip
 
-Because the sessions table builds from our `_events_this_run` table, and both the `condition` and `value` fields will accept any valid sql that can go in a `case when...` and `select` block respectively, this means you can use any fields in a `contexts_` or `unstruct_` column in addition to the default `atomic.events` fields. However, it will not accept dbt code so you need to extract the relevant field yourself. You can see an example of this below.
+Because the sessions table builds from our `_events_this_run` table, and both the `condition` and `value` fields will accept any valid sql that can go in a `case when...` and `select` block respectively, this means you can use any fields in a `contexts_` or `unstruct_` column in addition to the default `atomic.events` fields. However, it will not accept dbt code so you need to extract the relevant field yourself. You can see an example of this below.
```
emielver, line 72:
```suggestion
Because the sessions table builds from our `_events_this_run` table, and both the `condition` and `value` fields will accept any valid sql that can go in a `case when...` and `select` block respectively, this means you can use any fields in a `contexts_` or `unstruct_` column in addition to the default `atomic.events` fields. However, it will not accept dbt code so you need to extract the relevant field yourself. You can see an example of this below.
```

PR #655, opened 2023-10-25
Add accelerators link to header

Jack-Keene:
This PR adds a link to the accelerators on the page header. 

Q: Do we need this link to the docs in the header? 

image

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -115,7 +115,6 @@ module.exports = {
             label: 'GitHub',
             position: 'left',
           },
-
           {
             to: 'https://try.snowplowanalytics.com/?utm_content=hero-cta&utm_campaign=snowplow-docs',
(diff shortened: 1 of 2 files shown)


docusaurus.config.js
```diff
@@ -110,6 +115,7 @@ module.exports = {
             label: 'GitHub',
             position: 'left',
           },
+
```
greg-el, line 118: 👀

greg-el:
Looks like it gets a bit funky when the page is thinner
image
image

greg-el:
Reguarding your Q, I reckon we can get rid of that "Docs" link, it seems pretty redundent since the "SNOWPLOW" logo goes to the docs home too which is a fairly standard convention. @stanch what do you think?

greg-el (approved):
LGTM

PR #668, opened 2023-11-14
Fix schema gen

Jack-Keene:
This PR rewrites the state logic for versioned schemas for the docs. It also simplifies the return statements and makes the code more readable.

There's still quite a few improvements i'd like to make on this but I think this is a good start


rlh1994:
When you remove the selected schema, it still renders the schema setter elements, otherwise it looks good!
image

PR #686, opened 2023-11-24
Bump versions and add unified to actions

rlh1994:
Bump a few versions we've missed recently, and adds unified to the actions that calculate all the versions stuff


georgewoodhead (approved):
lgtm!

PR #699, opened 2023-11-29
Add docs for OTB button click plugin

greg-el:


docs/collecting-data/collecting-from-own-applications/javascript-trackers/web-tracker/plugins/button-click/index.md
```diff
@@ -0,0 +1,339 @@
...
+```
+
+</TabItem>
+</Tabs>
+
+## Configuration
+
+### Choosing which buttons to track
+
+There are three ways to configure which buttons are tracked:
```
matus-tomlein, line 150: You could say there are four ways to configure – one being without configuration. I think it's worth mentioning what is tracked when no configuration is provided.

matus-tomlein (approved):
LGTM!

PR #711, opened 2023-12-07
Update default value for the enable_mobile variable in unified

emmahunt:
This PR updates the default value in the documentation, which is incorrect - the mobile data is enabled by default


rlh1994 (approved):
The hero we need

PR #717, opened 2023-12-13
Update instructions for using React Native tracker both in React Native and native code

matus-tomlein:
Updates a section in React Native tracker docs for accessing the tracker from native code.


greg-el (approved):
LGTM!

PR #722, opened 2023-12-19
Update config generator package name from web to unified

emmahunt:
The config generator on the [unified](https://docs.snowplow.io/docs/modeling-your-data/modeling-your-data-with-dbt/dbt-configuration/unified/) page is set to web: this PR updates it to unified (I'm not certain this is the right way but the preview build looks correct)


agnessnowplow (approved):
LGTM! Thanks for updating this.

PR #732, opened 2024-01-11
Fix outdated variable argument names

rlh1994:
This still uses the old `name` value in place, it should be `schema` everywhere


georgewoodhead (approved):
lgtm!

PR #741, opened 2024-01-19
Bump versions, license, schema for ecom 0.7.0

rlh1994:


agnessnowplow (approved):
Just one nit, I would group the warehouse and tracker related vars one after the other (basically moving the category separator below):
image

PR #742, opened 2024-01-19
Remove pagination and add quick links at top of page

rlh1994:
Using pagination means it can be hard to search for the variable you want on the page, especially as we split them into different tables. This ensures all variables are visible and loading on the page.

As this can lead to quite long pages, I had added quick links to the top of the page - note it is _technically_ possible to force a true sidebar table of contents on the page, I think these pages benefit from the extra width so took this approach.


georgewoodhead:
The links don't seem to quite take you all the way to the relevant section?
Screen Recording 2024-01-23 at 09 35 44

georgewoodhead (approved):

PR #749, opened 2024-01-23
Enrich 4.0.0

benjben:
- [x] Add info about fields' limits in configuration reference
- [x] Add Azure info to `enrich-kafka`'s configuration reference


docs/pipeline-components-and-applications/enrichment-components/4-0-x-upgrade-guide/index.md
```diff
@@ -0,0 +1,62 @@
...
+}
+```
+
+Default limits are equal to the one set in atomic iglucentral [schema](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/atomic/jsonschema/1-0-0), apart from:
+
+* `mkt_clickid` limit increased from 128 to 1000.
+* `page_url` limit increased from 4096 to 10000.
+* `page_referrer` limit increased from 4096 to 10000.
+
+Therefore, atomic fields with longer values could break downstream loading to data warehouse (Redshift/Snowflake/Databricks).
```
stanch, line 55: In my understanding, Redshift would not break, because Redshift itself truncates the values. Also, we should detail what happens depending on the setting of `accept_invalid`

docs/pipeline-components-and-applications/enrichment-components/4-0-x-upgrade-guide/index.md
```diff
@@ -0,0 +1,62 @@
...
+
+To avoid any issues with loading, we recommend either:
+
+1) Remove (or increase) limits defined on VARCHAR-like columns in data warehouse. (#TODO add more specific queries for each destination?)
```
stanch, line 59: We should recommend:
* For Redshift — increase the VARCHAR columns to the new defaults (optional, otherwise they will be truncated)
* For Snowflake — remove the limit, as Snowflake says there is no difference
* For Databricks — same as Snowflake

docs/pipeline-components-and-applications/enrichment-components/4-0-x-upgrade-guide/index.md
```diff
@@ -0,0 +1,62 @@
...
+OR 
+2) Set `truncateAtomicFields` (#TODO docs link?) value in RDB Transformer configuration. That would result in truncated values for fields exceeding limits. As opposed to Enrich, RDB Transformer relies on limits defined in atomic iglucentral schema.
```
stanch, line 61: I actually think we should not be recommending this option, because it relies on an experimental flag, and only makes sense in the short term.

For the second alternatively, I would instead suggest to change the new defaults back to their original values.

docs/pipeline-components-and-applications/enrichment-components/configuration-reference/index.md
```diff
@@ -23,6 +23,7 @@ sidebar_position: 50
 | `telemetry.userProvidedId` | Optional. See [here](/docs/getting-started-on-community-edition/telemetry/index.md#how-can-i-help) for more information. |
 | `featureFlags.acceptInvalid` | Optional. Default: `false`. Enrich *3.0.0* introduces the validation of the enriched events against atomic schema before emitting. If set to `false`, a bad row will be emitted instead of the enriched event if validation fails. If set to `true`, invalid enriched events will be emitted, as before. |
 | `featureFlags.legacyEnrichmentOrder` | Optional. Default: `false`. In early versions of `enrich-kinesis` and `enrich-pubsub` (>= *3.1.5*), the Javascript enrichment incorrectly ran before the currency, weather, and IP Lookups enrichments. Set this flag to true to keep the erroneous behavior of those previous versions. |
+| `validation.atomicFieldsLimits`      | Optional. Default: (#TODO Link to `application.conf` with defaults in repo...?) Configuration for custom maximum atomic fields (strings) length. It's a map-like structure with keys being atomic field names and values being their max allowed length.                                                 |
```
stanch, line 26:
```suggestion
| `validation.atomicFieldsLimits`      | Optional. For the defaults, see [here](https://github.com/snowplow/enrich/blob/master/modules/common/src/main/resources/reference.conf). Configuration for custom maximum atomic fields (strings) length. It's a map-like structure with keys being atomic field names and values being their max allowed length.                                                 |
```

docs/pipeline-components-and-applications/enrichment-components/4-0-x-upgrade-guide/index.md
```diff
@@ -0,0 +1,62 @@
+---
+title: "4.0.x upgrade guide"
+sidebar_position: 0
+---
+
+## Breaking changes
```
stanch, line 6: Should we also mention that the experimental rabbitmq artifact is no longer published?

stanch (changes requested):

docs/pipeline-components-and-applications/enrichment-components/4-0-x-upgrade-guide/index.md
benjben: Yeah that doesn't hurt to mention it

docs/pipeline-components-and-applications/enrichment-components/configuration-reference/index.md
```diff
@@ -23,6 +35,7 @@ sidebar_position: 50
 | `telemetry.userProvidedId` | Optional. See [here](/docs/getting-started-on-community-edition/telemetry/index.md#how-can-i-help) for more information. |
 | `featureFlags.acceptInvalid` | Optional. Default: `false`. Enrich *3.0.0* introduces the validation of the enriched events against atomic schema before emitting. If set to `false`, a bad row will be emitted instead of the enriched event if validation fails. If set to `true`, invalid enriched events will be emitted, as before. |
 | `featureFlags.legacyEnrichmentOrder` | Optional. Default: `false`. In early versions of `enrich-kinesis` and `enrich-pubsub` (>= *3.1.5*), the Javascript enrichment incorrectly ran before the currency, weather, and IP Lookups enrichments. Set this flag to true to keep the erroneous behavior of those previous versions. |
+| `validation.atomicFieldsLimits` | Optional. For the defaults, see [here](https://github.com/snowplow/enrich/blob/master/modules/common/src/main/resources/reference.conf). Configuration for custom maximum atomic fields (strings) length. It's a map-like structure with keys being atomic field names and values being their max allowed length. |
```
stanch, line 38: Could you add `_(since 4.0.0)_`?

docs/pipeline-components-and-applications/enrichment-components/configuration-reference/index.md
```diff
@@ -152,6 +165,15 @@ A minimal configuration file can be found on the [Github repo](https://github.co
 | `output.bad.topicName` | Optional. Name of the Kafka topic to write to |
 | `output.bad.bootstrapServers` | Optional. A list of host:port pairs to use for establishing the initial connection to the Kafka cluster |
 | `output.bad.producerConf` | Optional. Kafka producer configuration. See [the docs](https://kafka.apache.org/documentation/#producerconfigs) for all properties |
+| `blobStorage.s3` | Optional. Set to `true` if S3 client should be initialized to download enrichments assets. |
```
stanch, line 168: Same comment as above

docs/pipeline-components-and-applications/enrichment-components/configuration-reference/index.md
```diff
@@ -152,6 +165,15 @@ A minimal configuration file can be found on the [Github repo](https://github.co
...
+| `blobStorage.gcs` | Optional. Set to `true` if GCS client should be initialized to download enrichments assets. |
+| `blobStorage.azureStorage` | Optional. Azure Blob Storage client configuration. ABS client won't be enabled if it isn't given. |
+| `blobStorage.azureStorage.accounts` | Array of accounts to download from Azure Blob Storage. |
+
+Example values for the storage accounts :
```
stanch, line 173:
```suggestion
Example values for the Azure storage accounts :
```

docs/pipeline-components-and-applications/enrichment-components/configuration-reference/index.md
```diff
@@ -264,6 +220,8 @@ If `acceptInvalid` is set to `false`, a bad row will be emitted instead of the e
 
 When we'll know that all our customers don't have any invalid enriched events any more, we'll remove the feature flags and it will be impossible to emit invalid enriched events.
 
+Since `4.0.0`, it is possible to configure the lengths of the atomic fields with `validation.atomicFieldsLimits` (see above).
```
stanch, line 223: Worth copy-pasting the example here I think — the one from the “upgrading” page.

stanch (changes requested):

stanch (approved):

PR #755, opened 2024-01-29
Update JavaScript tracker docs to 3.21.0

igneel64:


matus-tomlein (approved):
LGTM! But please also update the component versions: https://github.com/snowplow/documentation/blob/main/src/componentVersions.js#L11

PR #774, opened 2024-02-12
Add images for the funnel builder app and make small changes in the app docs

matus-tomlein:
This PR extends the documentation for the Funnel Builder data app to:

1. Add screenshots of the app and relevant parts of the UI.
2. Make small improvements in the text.

If you see any other improvements we could make to the docs, please shout!


rlh1994:
Could we write it more like a step by step guide to building a funnel, and the UI screenshots are there to support those steps, instead of focusing on describing the UI which is feels a bit like at the moment. This helps as the UI goes out of date that it's still a relevant guide.

Also make sure all images have suitable alt text please

matus-tomlein:
Good point @rlh1994! Have now reworded it to be more like a step by step guide.

rlh1994 (approved):
Maybe move the screenshot of the whole app above the requirements header, but otherwise looks good!

PR #787, opened 2024-02-28
Add Snowflake Streaming Loader

stanch:


istreeter:
@stanch I don't think you mentioned the license anywhere.  Just pointing it out in case that was an oversight.

We already discussed this elsewhere, but I really don't like the name Snowflake Streaming Loader.

docs/pipeline-components-and-applications/loaders-storage-targets/snowflake-streaming-loader/configuration-reference/_kafka_config.md
```diff
@@ -0,0 +1,28 @@
+<tr>
```
istreeter, line 1: For all our common-streams apps, these kinesis/kafka/pubsub configuration inclusions are going to be the same everywhere.  One day when I have more energy, I might split them out to a common re-usable block in the docs site too.

docs/pipeline-components-and-applications/loaders-storage-targets/snowflake-streaming-loader/configuration-reference/_common_config.md
```diff
@@ -0,0 +1,97 @@
...
+    <td>Optional. Sets the login timeout on the JDBC driver which connects to Snowflake</td>
+</tr>
+<tr>
+    <td><code>output.good.jdbcNetworkTimeout</code></td>
+    <td>Optional. Sets the network timeout on the JDBC driver which connects to Snowflake</td>
+</tr>
+<tr>
+    <td><code>output.good.jdbcQueryTimeout</code></td>
+    <td>Optional. Sets the query timeout on the JDBC driver which connects to Snowflake</td>
+</tr>
```
istreeter, line 97: Just pointing out these `output.good.jdbc*Timeout`s should really belong in the Snowflake configuration section.  But they are of such low importance that I decided to move them to a less valuable part of the page.

istreeter (approved):

PR #795, opened 2024-03-07
Update Self Describing Event example js call

domfrecent:
Remove `event` as the highest level object in the example `window.snowplow()` javascript call for`'trackSelfDescribingEvent'`. The collector is not able to interpret the call with event{}

### DIFF
--- a/docs/understanding-your-pipeline/events/index.md
+++ b/docs/understanding-your-pipeline/events/index.md
@@ -147,18 +147,16 @@ Then you can use one of our [tracking SDKs](/docs/collecting-data/collecting-fro
 
 ```javascript
 window.snowplow('trackSelfDescribingEvent', {
-  event: {
-    schema: 'iglu:com.acme_company/viewed_product/jsonschema/2-0-0',
-    data: {
-      productId: 'ASO01043',
-      category: 'Dresses',
-      brand: 'ACME',
-      returning: true,
-      price: 49.95,
-      sizes: ['xs', 's', 'l', 'xl', 'xxl'],
-      availableSince: new Date(2013,3,7)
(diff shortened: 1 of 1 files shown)


snowplowcla:
Thanks for your pull request. Is this your first contribution to a Snowplow open source project? Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).

:memo: Please visit https://docs.snowplowanalytics.com/docs/contributing/contributor-license-agreement/ to learn more and sign.

Once you've signed, please reply here (e.g. I signed it!) and we'll verify. Thanks.

domfrecent:
I signed CLA!

snowplowcla:
Confirmed! @domfrecent has signed the Contributor License Agreement. Thanks so much.

matus-tomlein:
Hi @domfrecent, thank you for the contribution, but I think it's necessary to embed the `schema` and `data` under the `event` key. That is also what the [docs under the JS tracker](https://docs.snowplow.io/docs/collecting-data/collecting-from-own-applications/javascript-trackers/web-tracker/custom-tracking-using-schemas/?platform=js#tracking-a-custom-event-self-describing) section suggest and when I try this, I get this following error:

Screenshot 2024-03-29 at 08 41 57

I think the problem might be that you are using an old version of the JS tracker – version 2 or below. In that version, the `event` key was not necessary. However, this is a very old version that is not supported anymore and it's highly advisable to upgrade to version 3 of the tracker.

domfrecent:
Hi @matus-tomlein thanks for your response! You're absolutely right, I was on a 2.X version of the tracker and was running into syntactic inconsistencies because of that. 

We can close this PR - may be helpful to add an indication in the documentation somewhere for the version that the documentation reflects.

PR #801, opened 2024-03-13
Add schema validation

rlh1994:
“Your scientists were so preoccupied with whether they could, they didn’t stop to think if they should.”


src/components/JsonSchemaValidator/dbt-variables-meta-schema
```diff
@@ -0,0 +1,47 @@
+{
+    "$id": "dbt-variables-meta-schema",
```
jbeemster, line 2: Great work shrinking this down to just required changes!

jbeemster (approved):
Schemas validate and meta-schema is readable!  Haven't checked anything beyond that but I trust you to have done your own due diligence on how this behaves on the site.

PR #815, opened 2024-03-25
Refactor Enterprise & Cloud

stanch:
* Simplified tiering in the docs (BDP vs CE, instead of Enterprise, Cloud and CE)
* Removed most explicit references to Cloud/Enterprise due to functional parity
* Removed outdated Cloud content
* Cleaned up the setup pages a little


ebedford420 (approved):
looks good, thank you :)

PR #819, opened 2024-03-27
Add Event Specifications plugin

igneel64:
Add the Event Specifications plugin to documentation

_Updated version to 3.23.0 which is expected to be released with the plugin_


docs/collecting-data/collecting-from-own-applications/javascript-trackers/web-tracker/tracking-events/event-specifications/index.md
```diff
@@ -0,0 +1,79 @@
...
+```
+
+  </TabItem>
+</Tabs>
+
+## Configuration
+
+The configuration of the plugin would be in the form of:
+
+```ts
```
matus-tomlein, line 68: Can we also give instructions how to configure the plugin when using the JS tag? It's not super intuitive, there is a [general guide here](https://docs.snowplow.io/docs/collecting-data/collecting-from-own-applications/javascript-trackers/web-tracker/plugins/configuring-tracker-plugins/javascript/#plugins-after-initialization), but better to have it as part of the plugin docs I think.

matus-tomlein (approved):
LGTM!

PR #821, opened 2024-03-29
Single table for events table

pkutaj:
* Discussion: https://snowplow.slack.com/archives/CGK462GG3/p1711710007601719


jethron:
- What's the order used here? I would expect it to match the enriched TSV order as defined in e.g. [Redshift's atomic.events](https://github.com/snowplow/snowplow/blob/master/4-storage/redshift-storage/sql/atomic-def.sql), [an Analytics SDK](https://github.com/snowplow/snowplow-python-analytics-sdk/blob/master/snowplow_analytics_sdk/event_transformer.py), [BQ Loader](https://github.com/snowplow-incubator/snowplow-bigquery-loader/blob/master/modules/mutator/src/main/scala/com/snowplowanalytics/snowplow/storage/bigquery/mutator/Atomic.scala), etc.
- (along these lines this is also missing `etl_tags` - which to be fair is useless - but is there)
- `ti_base_price` seems to have some note attached that's now undefined
- For this kind of reference I can imagine the field length limits being useful with a reference to [`validation.atomicFieldsLimits`](https://docs.snowplow.io/docs/pipeline-components-and-applications/enrichment-components/configuration-reference/#common-parameters) in case it becomes an issue; but I can appreciate that will be very annoying to add so definitely not a blocker by any means

PR #839, opened 2024-04-05
Clean up data modeling section

stanch:
* Tweaked the sidebar labels
* Moved Analytics SDKs to streaming data elsewhere, which is where they belong


docs/destinations/analytics-sdk/index.md
```diff
@@ -0,0 +1,24 @@
+---
```
stanch, line 1: Moved only

rlh1994 (approved):
sure seems sensible

PR #877, opened 2024-05-07
Add BigQuery Loader V2

pondzix:


docs/storing-querying/loading-process/index.md
```diff
@@ -25,11 +26,14 @@ We load data into Redshift using the [RDB Loader](/docs/pipeline-components-and-
 
   </TabItem>
   <TabItem value="bigquery" label="BigQuery">
-
-We load data into BigQuery using the [BigQuery Loader](/docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/index.md).
-
-<BigQueryLoaderDiagram/>
-
+    <Tabs groupId="bqgroup" queryString lazy>
+        <TabItem value="v2" label="2.x" default>
```
istreeter, line 30: Hmmm.... I'd probably go for "Version 2.x" instead of "2.x".  To make it more clear what the button refers to.

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/upgrade-guides/2-0-0-upgrade-guide/index.md
```diff
@@ -0,0 +1,22 @@
...
+
+BigQuery Loader 2.0.0 brings changes to the loading setup. It is no longer neccessary to configure and deploy three (in [1.X](/docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/previous-versions/bigquery-loader-1.x/index.md) it was loader, repeater and mutator) independent applications in order to load your data to BigQuery.
+Starting from 2.0.0 only one appliction is needed, which naturally introduces some breaking changes to the configuration file structure.
+
+See the [configuration reference](/docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/configuration-reference/index.md) for all possible configuration parameters
+and the minimal [configuration samples](https://github.com/snowplow-incubator/snowplow-bigquery-loader/blob/v2/config) for each of supported cloud environments.
+
+## Events table format
+
+Starting from 2.0.0, BigQuery Loader no longer uses full schema version in column names for self-describing events and entities in `events` table. It uses only major schema version in the column name instead.
```
istreeter, line 16: I suggest drop in an example here.  Pick a commonly used schema, and say what the column is called before and after the upgrade.  Will be much quicker for the user to grasp, rather than explaining it with words.

After that, I think you can condense the first two paragraphs into fewer words.

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/upgrade-guides/2-0-0-upgrade-guide/index.md
```diff
@@ -0,0 +1,22 @@
...
+
+This means all new self-describing events and entities will be loaded to 'major version' - oriented columns, whereas old 'full version' - oriented columns remain unchanged, are no longer used by the new loader, and have no direct effect on loading. If neccessary, old columns need to be handled/consolidated separately. //TODO how?
```
istreeter, line 18: The point about consolidating old columns is important.  You can ask our analytics engineers for help with phrasing this advice.

I _think_ the advice is..... if you use Snowplow's standard DBT models then don't worry it will all work ok automatically.  If you don't use Snowplow's DBT models then go figure it out for yourselves.

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/upgrade-guides/2-0-0-upgrade-guide/index.md
```diff
@@ -0,0 +1,22 @@
+---
+title: "2.0.0 upgrade guide"
+sidebar_position: -20
+---
+
+## Configuration
+
+BigQuery Loader 2.0.0 brings changes to the loading setup. It is no longer neccessary to configure and deploy three (in [1.X](/docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/previous-versions/bigquery-loader-1.x/index.md) it was loader, repeater and mutator) independent applications in order to load your data to BigQuery.
```
istreeter, line 8: Beyond just configuration... we also make obsolete some of the cloud infrastructure.  E.g. the pubsub topic that connected the 3 components.  And the GCS bucket where the repeater loaded failed inserts.

I think we can re-write this paragraph to improve readability.  But please add that extra info first, and then we can improve readability after we've got the content.

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/_cross-cloud-diagram.md
```diff
@@ -0,0 +1,18 @@
...
+
+<Tabs groupId="cloud" queryString lazy>
+  <TabItem value="aws" label="AWS">
+    <LoaderDiagram {...props} stream="Kinesis" cloud="AWS"/>
+  </TabItem>
+  <TabItem value="gcp" label="GCP">
+    <LoaderDiagram {...props} stream="Pub/Sub" cloud="GCP"/>
+  </TabItem>
+  <TabItem value="azure" label="Azure">
+    <LoaderDiagram {...props} stream="Kafka"cloud="Azure"/>
```
stanch, line 16:
```suggestion
    <LoaderDiagram {...props} stream="Kafka" cloud="Azure"/>
```

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/_cross-cloud-diagram.md
```diff
@@ -0,0 +1,18 @@
...
+```
+
+<Tabs groupId="cloud" queryString lazy>
+  <TabItem value="aws" label="AWS">
+    <LoaderDiagram {...props} stream="Kinesis" cloud="AWS"/>
+  </TabItem>
+  <TabItem value="gcp" label="GCP">
+    <LoaderDiagram {...props} stream="Pub/Sub" cloud="GCP"/>
+  </TabItem>
+  <TabItem value="azure" label="Azure">
```
stanch, line 15: Do we actually have an Azure artifact?

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/index.md
```diff
@@ -1,316 +1,64 @@
...
+    <LoaderDiagram stream="Pub/Sub" cloud="GCP"/>
+    <DeployOverview cloud="GCP" stream="pubsub"/>
+  </TabItem>
+  <TabItem value="azure" label="Azure">
+    <LoaderDiagram stream="Kafka" cloud="Azure"/>
+    <DeployOverview cloud="Azure" stream="kafka"/>
+  </TabItem>
+</Tabs>
 
-For more information on how events are stored in BigQuery, check the [mapping between Snowplow schemas and the corresponding BigQuery column types](/docs/storing-querying/schemas-in-warehouse/index.md?warehouse=bigquery).
```
stanch, line 18: Could you keep this tip somewhere on this page?

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/previous-versions/bigquery-loader-1.x/_diagram.md
```diff
@@ -0,0 +1,24 @@
+At the high level, BigQuery loader reads enriched Snowplow events in real time and loads them in BigQuery using the Storage Write API.
```
stanch, line 1: Doesn’t 1.x use Streaming Inserts API rather than Storage Write API?

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/previous-versions/bigquery-loader-1.x/index.md
```diff
@@ -0,0 +1,315 @@
...
+  "monitoring": {} # disabled
+}
+```
+
+The loader takes command line arguments `--config` with a path to the configuration hocon file and `--resolver` with a path to the Iglu resolver file. If you are running the docker image then you should mount the configuration files into the container:
+
+<CodeBlock language="bash">{
+`docker run \\
+    -v /path/to/configs:/configs \\
+    snowplow/snowplow-bigquery-streamloader:${versions.bqLoader} \\
```
stanch, line 150: All snippets on this page should use the latest 1.x version rather than `versions.bqLoader`.

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/upgrade-guides/1-0-x-upgrade-guide/index.md
```diff
@@ -6,7 +6,7 @@ sidebar_position: 0
 
 ## Configuration
 
-The only breaking change from the 0.6.x series is the new format of the configuration file. That used to be a self-describing JSON but is now HOCON. Additionally, some app-specific command-line arguments have been incorporated into the config, such as Repeater's `--failedInsertsSub` option. For more details, see the [setup guide](/docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/index.md#setup-guide) and [configuration reference](/docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/snowplow-bigquery-loader-configuration-reference/index.md).
+The only breaking change from the 0.6.x series is the new format of the configuration file. That used to be a self-describing JSON but is now HOCON. Additionally, some app-specific command-line arguments have been incorporated into the config, such as Repeater's `--failedInsertsSub` option. For more details, see the [setup guide](/docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/index.md#setup-guide) and
```
stanch, line 9: and...?

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/upgrade-guides/2-0-0-upgrade-guide/index.md
stanch, line 8:
```suggestion
BigQuery Loader 2.0.0 brings changes to the loading setup. It is no longer neccessary to configure and deploy three independent applications (Loader, Repeator and Mutator in [1.X](/docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/previous-versions/bigquery-loader-1.x/index.md)) in order to load your data to BigQuery.
```

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/upgrade-guides/2-0-0-upgrade-guide/index.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Events table format
+
+Starting from 2.0.0, BigQuery Loader no longer uses full schema version in column names for self-describing events and entities in `events` table. It uses only major schema version in the column name instead.
+
+This means all new self-describing events and entities will be loaded to 'major version' - oriented columns, whereas old 'full version' - oriented columns remain unchanged, are no longer used by the new loader, and have no direct effect on loading. If neccessary, old columns need to be handled/consolidated separately. //TODO how?
+
+Before 2.0.0, breaking changes introduced for the same schema family (shemas using the same major version) had no impact on your `events` table. Starting from 2.0.0, loader tries to merge all changes from the same schema family and load to the single column (with major version suffix). In case of breaking changes, loader creates recovery columns to try to load all your data, even the one referencing 'broken' schemas. You can read more about schema evolution and how recovery columns work [here](/docs/storing-querying/schemas-in-warehouse/?warehouse=bigquery#versioning).
```
stanch, line 20: I think this needs a concrete example, similar to what we had in Discourse announcements for recovery columns/tables in Databricks/Redshift

docs/storing-querying/loading-process/index.md
```diff
@@ -25,11 +26,14 @@ We load data into Redshift using the [RDB Loader](/docs/pipeline-components-and-
 
   </TabItem>
   <TabItem value="bigquery" label="BigQuery">
-
-We load data into BigQuery using the [BigQuery Loader](/docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/index.md).
-
-<BigQueryLoaderDiagram/>
-
+    <Tabs groupId="bqgroup" queryString lazy>
```
stanch, line 29:
```suggestion
    <Tabs groupId="biquery-loader-version" queryString lazy>
```

docs/storing-querying/querying-data/index.md
```diff
@@ -80,12 +80,16 @@ You can query fields in the self-describing event like so:
...
 SELECT
     ...
-    unstruct_event_my_example_event_1_0_0.my_field,
+    unstruct_event_my_example_event_1.my_field,
     ...
 FROM
     <events>
 ```
 
+:::caution
```
stanch, line 89: I think this is a note and not a caution

docs/storing-querying/querying-data/index.md
```diff
@@ -190,8 +194,11 @@ SELECT
 FROM
     <events>
 LEFT JOIN
-    unnest(contexts_my_entity_1_0_0) AS my_ent -- left join to avoid discarding events without values in this entity
+    unnest(contexts_my_entity_1) AS my_ent -- left join to avoid discarding events without values in this entity
 ```
+:::caution
```
stanch, line 199: Ditto

docs/storing-querying/schemas-in-warehouse/index.md
```diff
@@ -108,7 +108,57 @@ For example, suppose you have the following field in the schema:
 It will be translated into a field called `last_name` (notice the underscore), of type `STRING`.
 
   </TabItem>
+
   <TabItem value="bigquery" label="BigQuery">
+    <Tabs groupId="bqgroup" queryString lazy>
```
stanch, line 113: Could you put just the differing paragraphs in tabs?

stanch (changes requested):

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/_cross-cloud-diagram.md
pondzix: Does `snowplow/bigquery-loader-kafka` count as Azure artifact? 😁

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/configuration-reference/_bigquery_config.md
```diff
@@ -0,0 +1,16 @@
+<tr>
```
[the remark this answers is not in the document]
pondzix: Yes, it was hard copy&paste work!

Now I noticed... shouldn't sentry be under `monitoring` section?

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/previous-versions/bigquery-loader-1.x/_diagram.md
pondzix: Yeah...something like this. I'm confused by the naming. In docs I think they call it [legacy streaming API](https://cloud.google.com/bigquery/docs/streaming-data-into-bigquery). But the new one is also streaming, right?

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/_diagram.md
```diff
@@ -1,24 +1,16 @@
...
-    table[("<b>Events table</b>")]
-  end
-  stream-->loader-->BigQuery
+```mdx-code-block
+import Mermaid from '@theme/Mermaid';
+import Link from '@docusaurus/Link';
 ```
 
-BigQuery loader consists of three applications: Loader, Mutator and Repeater. The following diagram illustrates the interaction between them and BigQuery:
+<p>The BigQuery Streaming Loader on {props.cloud} is a fully streaming application that continually pulls events from {props.stream} and writes to BigQuery using the <Link to="https://cloud.google.com/bigquery/docs/reference/storage/libraries#client-libraries-install-java">BigQuery Storage API</Link></p>
```
stanch, line 6:
```suggestion
<p>The BigQuery Streaming Loader on {props.cloud} is a fully streaming application that continually pulls events from {props.stream} and writes to BigQuery using the <Link to="https://cloud.google.com/bigquery/docs/reference/storage/libraries#client-libraries-install-java">BigQuery Storage API</Link>.</p>
```

docs/storing-querying/loading-process/index.md
```diff
@@ -25,11 +26,14 @@ We load data into Redshift using the [RDB Loader](/docs/pipeline-components-and-
 
   </TabItem>
   <TabItem value="bigquery" label="BigQuery">
-
-We load data into BigQuery using the [BigQuery Loader](/docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/index.md).
```
stanch, line 29: Could you bring back this line? It’s useful as a link to the BQ Loader page, and also to explain what Versions 2.x/1.x refer to.

docs/storing-querying/schemas-in-warehouse/index.md
```diff
@@ -332,7 +351,24 @@ Note that this behavior was introduced in RDB Loader 5.3.0.
...
+| `com.example/button_press/jsonschema/1-2-0` | `unstruct_event_com_example_button_press_1` |
+| `com.example/button_press/jsonschema/2-0-0` | `unstruct_event_com_example_button_press_2` |
 
+When you evolve your schema within the same major version, (non-destructive) changes are applied to the existing column automatically. For example, if you add a new optional field in the schema, a new optional field will be added to the `RECORD`.
+
+:::info Breaking changes
+<ParquetRecoveryColumns/>
+:::
+   </TabItem>
+   <TabItem value="v1" label="Version 1.x">
```
stanch, line 371: Could you adjust the text below? (See emphasized parts)

---

While our recommendation is to use major schema versions to indicate breaking changes (e.g. changing a type of a field from a `string` to a `number`), this is not particularly relevant with BigQuery **Loader version 1.x**. Indeed, each schema version gets its own column, so there is no difference between major and minor versions. That said,  sticking to our recommendation is a good idea:
 * Breaking changes might affect downstream consumers of the data, even if they don’t affect BigQuery
 * **Version 2 of the loader has stricter behavior that matches our loaders for other warehouses and lakes**

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/previous-versions/bigquery-loader-1.x/index.md
```diff
@@ -0,0 +1,315 @@
+---
+title: "BigQuery Loader (1.x)"
+sidebar_position: 0 
+---
+
+```mdx-code-block
+import {versions} from '@site/src/componentVersions';
+import CodeBlock from '@theme/CodeBlock';
+import Diagram from '@site/docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/_diagram.md';
```
stanch, line 9: This should be using the old diagram

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/previous-versions/bigquery-loader-1.x/index.md
```diff
@@ -0,0 +1,315 @@
...
+  "monitoring": {} # disabled
+}
+```
+
+The loader takes command line arguments `--config` with a path to the configuration hocon file and `--resolver` with a path to the Iglu resolver file. If you are running the docker image then you should mount the configuration files into the container:
+
+<CodeBlock language="bash">{
+`docker run \\
+    -v /path/to/configs:/configs \\
+    snowplow/snowplow-bigquery-streamloader:1.7.1 \\
```
stanch, line 150: I recommend creating a new variable like `versions.bqLoader1x` in case we release 1.7.2 in the meantime and need to adjust this.

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/index.md
```diff
@@ -1,316 +1,68 @@
...
+import LoaderDiagram from '@site/docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/_diagram.md';
+import DeployOverview from '@site/docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/_deploy-overview.md';
 ```
 
-Under the umbrella of Snowplow BigQuery Loader, we have a family of applications that can be used to load enriched Snowplow data into BigQuery.
-
-<Diagram/>
+## Overview
+
+The BigQuery Loader is an application that loads Snowplow events to BigQuery using the [BigQuery Storage API](https://cloud.google.com/bigquery/docs/write-api).
```
stanch, line 17: Could we avoid repetition here?

image

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/upgrade-guides/2-0-0-upgrade-guide/index.md
```diff
@@ -0,0 +1,118 @@
+---
+title: "2.0.0 upgrade guide"
+sidebar_position: -20
+---
+
+## Configuration
+
+BigQuery Loader 2.0.0 brings changes to the loading setup. It is no longer neccessary to configure and deploy three independent applications (Loader, Repeater and Mutator in [1.X](/docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/previous-versions/bigquery-loader-1.x/index.md)) in order to load your data to BigQuery.
+Starting from 2.0.0 only one appliction is needed, which naturally introduces some breaking changes to the configuration file structure.
```
stanch, line 9:
```suggestion
Starting from 2.0.0, only one appliction is needed, which naturally introduces some breaking changes to the configuration file structure.
```

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/upgrade-guides/2-0-0-upgrade-guide/index.md
```diff
@@ -0,0 +1,118 @@
...
+
+See the [configuration reference](/docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/configuration-reference/index.md) for all possible configuration parameters
+and the minimal [configuration samples](https://github.com/snowplow-incubator/snowplow-bigquery-loader/blob/v2/config) for each of supported cloud environments.
+
+## Infrastructure
+
+Apart from Repeater and Mutator, there are also other infrastructure components that have become obsolete:
```
stanch, line 16:
```suggestion
Apart from Repeater and Mutator, other infrastructure components have become obsolete:
```

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/upgrade-guides/2-0-0-upgrade-guide/index.md
```diff
@@ -0,0 +1,118 @@
...
+* The `types` PubSub topic connecting Loader and Mutator.
+* The `failedInserts` PubSub topic connecting Loader and Repeater.
+* The `deadLetter` GCS bucket used by Repeater to store data that repeatedly failed to be inserted into BigQuery.
+
+## Events table format
+
+Starting from 2.0.0, BigQuery Loader changes its output column naming strategy. For example, for [ad_click event](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow.media/ad_click_event/jsonschema/1-0-0):
+
+* Before an upgrade, new column would be named `unstruct_event_com_snowplowanalytics_snowplow_media_ad_click_event_1-0-0`.
```
stanch, line 25:
```suggestion
* Before an upgrade, the corresponding column would be named `unstruct_event_com_snowplowanalytics_snowplow_media_ad_click_event_1-0-0`.
```

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/upgrade-guides/2-0-0-upgrade-guide/index.md
```diff
@@ -0,0 +1,118 @@
...
+* After an upgrade, new column will be named `unstruct_event_com_snowplowanalytics_snowplow_media_ad_click_event_1`.
+
+This means all new self-describing events and entities will be loaded to 'major version' - oriented columns, whereas old 'full version' - oriented columns remain unchanged, are no longer used by the new loader, and have no direct effect on loading. If neccessary, old columns need to be consolidated separately. If you are [modeling your data with dbt](/docs/modeling-your-data/modeling-your-data-with-dbt/index.md), you can use [this macro](https://github.com/snowplow/dbt-snowplow-utils#combine_column_versions-source) to aggregate the data across multiple columns.
```
stanch, line 28:
```suggestion
All self-describing events and entities will be loaded to new 'major version' - oriented columns. Old 'full version' - oriented columns would remain unchanged, but no new data would be loaded into them. (The 2.0.0 loader just ignores these columns.)

The new column naming scheme has several advantages:
* Fewer columns created (BigQuery has a limit on the total number of columns)
* No need to update data models (or use complex macros) every time a new minor version of a schema is created

The catch is that you have to follow the rules of schema evolution more strictly to ensure data from different schema versions can fit in the same column — see below.

:::tip Consolidating old and new columns

If you are using [Snowplow dbt models](/docs/modeling-your-data/modeling-your-data-with-dbt/dbt-models/index.md), they will automatically consolidate the data between `_1_0_0` and `_1` style columns, because they look at the major version prefix (e.g. `_1`), which is common to both.

If you are not using Snowplow dbt models but still use dbt, you can employ [this macro](https://github.com/snowplow/dbt-snowplow-utils#combine_column_versions-source) to manually aggregate the data across old and new columns.

:::
```

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/upgrade-guides/2-0-0-upgrade-guide/index.md
```diff
@@ -0,0 +1,118 @@
...
+
+Loader tries to format the incoming data according to the latest version of the schema it saw (for a given major version, e.g. `1-*-*`). For example, if a batch contains events with schema versions `1-0-0`, `1-0-1` and `1-0-2`, the loader derives the output schema based on version `1-0-2`. Then the loader instructs BigQuery to adjust the database column and load the data.
+
+This logic relies on two assumptions:
+
+1. **Old events compatible with new schemas.** Events with older schema versions, e.g. `1-0-0` and `1-0-1`, have to be valid against the newer ones, e.g. `1-0-2`. Those that are valid will result in failed events.
+
+2. **Old columns compatible with new schemas.** The corresponding BigQuery columns have to be migrated correctly from one version to another. Changes, such as altering the type of a field from `integer` to `string`, would fail. Loading would break with SQL errors and the whole batch would be stuck and hard to recover.
+
+These assumptions are not always clear to the users, making the process error-prone.
```
stanch, line 52: This does not apply to BigQuery Loader. I guess it needs to be replaced with an explanation of why it was not relevant for Version 1.

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/upgrade-guides/2-0-0-upgrade-guide/index.md
```diff
@@ -0,0 +1,118 @@
...
+
+### With recovery columns
+
+First, we support schema evolution that’s not strictly backwards compatible (although we still recommend against it since it can confuse downstream consumers of the data). This is done by _merging_ multiple schemas so that both old and new events can coexist. For example, suppose we have these two schemas:
```
stanch, line 56: Does this logic apply here? Or only to Databricks?

stanch: Either way I think it’s not relevant to BigQuery as _everything_ was allowed before, including this.

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/upgrade-guides/2-0-0-upgrade-guide/index.md
```diff
@@ -0,0 +1,118 @@
...
+   // merged
+   "properties": {
+      "a": {"type": "integer"},
+      "b": {"type": "integer"}
+   }
+}
+```
+
+
+Second, the loader does not fail when it can’t modify the database column to store both old and new events. (As a reminder, an example would be changing the type of a field from `integer` to `string`.) Instead, it creates a _temporary_ column for the new data as an exception. The users can then run SQL statements to resolve this situation as they see fit. For instance, consider these two schemas:
```
stanch, line 88: This needs rewording — old loader would not fail either

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/upgrade-guides/2-0-0-upgrade-guide/index.md
```diff
@@ -0,0 +1,118 @@
...
+Because `1-0-1` events cannot be loaded into the same column with `1-0-0`, the data would be put in a separate column, e.g. `contexts_com_snowplowanalytics_ad_click_1_0_1_recovered_9999999`, where:
+  - `1_0_1` is the version of the offending schema;
+  - `9999999` is a hash code unique to the schema (i.e. it will change if the schema is overwritten with a different one).
+
+If you create a new schema `1-0-2` that reverts the offending changes and is again compatible with `1-0-0`, the data for events with that schema will be written to the original column as expected.
+
+### Notes
+
+- If events with incorrectly evolved schemas do not arrive, then the recovery column would not be created.
+- It is still possible to break loading by overwriting version `1-0-0` of the schema. Please, avoid doing that.
```
stanch, line 116:
```suggestion
:::note

If events with incorrectly evolved schemas never arrive, then the recovery column would not be created.

:::
```

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/upgrade-guides/2-0-0-upgrade-guide/index.md
```diff
@@ -0,0 +1,118 @@
...
+
+You can read more about schema evolution and how recovery columns work [here](/docs/storing-querying/schemas-in-warehouse/?warehouse=bigquery#versioning).
```
stanch, line 118:
```suggestion
You can read more about schema evolution and how recovery columns work [here](/docs/storing-querying/schemas-in-warehouse/index.md?warehouse=bigquery#versioning).
```

stanch (changes requested):

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/previous-versions/bigquery-loader-1.x/_diagram.md
istreeter: Let's be consistent with the GCP docs.  Change this sentence to say "legacy streaming API" and you can even link to that page you just shared.  And for the new one let's call it "storage write API" and [link to this](https://cloud.google.com/bigquery/docs/write-api).

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/upgrade-guides/2-0-0-upgrade-guide/index.md
```diff
@@ -0,0 +1,92 @@
...
+Apart from Repeater and Mutator, other infrastructure components have become obsolete:
+* The `types` PubSub topic connecting Loader and Mutator.
+* The `failedInserts` PubSub topic connecting Loader and Repeater.
+* The `deadLetter` GCS bucket used by Repeater to store data that repeatedly failed to be inserted into BigQuery.
+
+## Events table format
+
+Starting from 2.0.0, BigQuery Loader changes its output column naming strategy. For example, for [ad_click event](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow.media/ad_click_event/jsonschema/1-0-0):
+
+* Before an upgrade, the corresponding column would be named `unstruct_event_com_snowplowanalytics_snowplow_media_ad_click_event_1-0-0`.
```
stanch, line 25:
```suggestion
* Before an upgrade, the corresponding column would be named `unstruct_event_com_snowplowanalytics_snowplow_media_ad_click_event_1_0_0`.
```
Right?

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/upgrade-guides/2-0-0-upgrade-guide/index.md
```diff
@@ -0,0 +1,92 @@
...
+
+If you are not using Snowplow dbt models but still use dbt, you can employ [this macro](https://github.com/snowplow/dbt-snowplow-utils#combine_column_versions-source) to manually aggregate the data across old and new columns.
+
+:::
+
+## Recovery columns
+
+### What is schema evolution?
+
+One of Snowplow’s key features is the ability to [define custom schemas and validate events against them](https://docs.snowplow.io/docs/understanding-tracking-design/understanding-schemas-and-validation/). Over time, users often evolve the schemas, e.g. by adding new fields or changing existing fields. To accommodate these changes, BigQuery Loader 2.0.0 automatically adjusts the database tables in the warehouse accordingly.
```
stanch, line 48:
```suggestion
One of Snowplow’s key features is the ability to [define custom schemas and validate events against them](/docs/understanding-tracking-design/understanding-schemas-and-validation/index.md). Over time, users often evolve the schemas, e.g. by adding new fields or changing existing fields. To accommodate these changes, BigQuery Loader 2.0.0 automatically adjusts the database tables in the warehouse accordingly.
```

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/upgrade-guides/2-0-0-upgrade-guide/index.md
```diff
@@ -0,0 +1,92 @@
...
+
+There are two main types of schema changes:
+
+**Breaking**: The schema version has to be changed in a major way (`1-2-3` → `2-0-0`). As of BigQuery Loader 2.0.0, each major schema version has its own column (`..._1`, `..._2`, etc, for example: `contexts_com.snowplowanalytics_ad_click_1`).
+
+**Non-breaking**: The schema version can be changed in a minor way (`1-2-3` → `1-3-0` or `1-2-3` → `1-2-4`). Data is stored in the same database column.
+
+Loader tries to format the incoming data according to the latest version of the schema it saw (for a given major version, e.g. `1-*-*`). For example, if a batch contains events with schema versions `1-0-0`, `1-0-1` and `1-0-2`, the loader derives the output schema based on version `1-0-2`. Then the loader instructs BigQuery to adjust the database column and load the data.
```
stanch, line 56: I think this does not apply exactly like this because there are no batches.

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/upgrade-guides/2-0-0-upgrade-guide/index.md
```diff
@@ -0,0 +1,92 @@
...
+```json
+{
+   // 1-0-1
+   "properties": {
+      "a": {"type": "string"}
+   }
+}
+```
+
+Because `1-0-1` events cannot be loaded into the same column with `1-0-0`, the data would be put in a separate column, e.g. `contexts_com_snowplowanalytics_ad_click_1_0_1_recovered_9999999`, where:
```
stanch, line 80:
```suggestion
With BigQuery Loader 1.x, data for each version would go to its own column — no issue. With BigQuery Loader 2.x, there is only one column. But strings and integers can’t coexist!

To avoid crashing or losing data, BigQuery Loader 2.0.0 proceeds by creating a new column for the data with schema `1-0-1`, e.g. `contexts_com_snowplowanalytics_ad_click_1_0_1_recovered_9999999`, where:
```

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/upgrade-guides/2-0-0-upgrade-guide/index.md
```diff
@@ -0,0 +1,92 @@
...
+  - `1_0_1` is the version of the offending schema;
+  - `9999999` is a hash code unique to the schema (i.e. it will change if the schema is overwritten with a different one).
+
+If you create a new schema `1-0-2` that reverts the offending changes and is again compatible with `1-0-0`, the data for events with that schema will be written to the original column as expected.
+
```
stanch, line 85:
```suggestion
:::tip
You might find that some of your schemas were evolved incorrectly in the past, which results in the creation of these “recovery” columns after the upgrade. To address this for a given schema, create a new _minor_ schema version that reverts the breaking changes introduced in previous versions. (Or, if you want to keep the breaking change, create a new _major_ schema version.) You can set it to [supersede](/docs/understanding-tracking-design/versioning-your-data-structures/amending/index.md#marking-the-schema-as-superseded) the previous version(s), so that events are automatically validated against the new schema.
:::
```

Do we also need instructions similar to these? https://docs.snowplow.io/docs/pipeline-components-and-applications/loaders-storage-targets/snowplow-rdb-loader/upgrade-guides/6-0-0-upgrade-guide/#identifying-schemas-that-need-patching @istreeter

stanch (changes requested):
Could you also update this page? https://docs.snowplow.io/docs/storing-querying/storage-options/ BigQuery should be now an option for the 3 clouds

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/upgrade-guides/2-0-0-upgrade-guide/index.md
istreeter: That section "Identifying schemas that need patching" was written long ago before we had the supercededBy feature.  I would think now we have supercededBy, we shouldn't recommend to anyone to patch schemas.

Instead, the user can publish the fixed schema as a new major version, and configure the old schemas to automatically upgrade to the new major version.  This is consistent with our belief that schemas are immutable.  And it means we don't need to write a new igluctl feature.

stanch: My doubt was not specifically about patching but whether running igluctl would be useful to identify issues with schemas... Because I think it says what the incompatibilities are? With just a recovery column, how will the users know what to fix?

istreeter: igluctl has an existing sub-command `verify-parquet`

```
igluctl static verify-parquet <schema_dir>
```

It prints out warnings like

```
'com.test/test/jsonschema/1-0-1'. Changes: Incompatible type change String to Double at /id"
```

Unfortunately, it only works if the schemas are written into files on the local file system.  It does not connect to remote Iglus.

We've not really planned what happens when existing users migrate to the new loader.  And that is semi-deliberate: it is a lot of work to gather all the requirements and plan for all the things that could go wrong.  So I've been delaying it until there is time to work on it properly.

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/upgrade-guides/2-0-0-upgrade-guide/index.md
pondzix: There are batches, but more kind of... micro-batches? And within those micro-batches loader derives output schema.

docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/upgrade-guides/2-0-0-upgrade-guide/index.md
pondzix: @stanch @istreeter as BQ Loader v2 relies on the same schema compatibility model/rules as parquet/Databricks/lake loader, we could use `verify-parquet` igluctl command and recommend it to users to detect incompatibilities. But this `parquet` in the name is confusing! Should we also do some renaming in igluctl then to make it more generic? And then also adjust [docs](https://deploy-preview-877--snowplow-docs.netlify.app/docs/pipeline-components-and-applications/iglu/igluctl-2/#igluctl)

docs/storing-querying/storage-options/index.md
```diff
@@ -25,6 +25,7 @@ The cloud selection is for where your Snowplow pipeline runs. The warehouse itse
 | Destination | Type | Loader application | Status |
 | --- | --- | --- | --- |
 | Redshift<br/>_(including Redshift serverless)_ | Batching (recommended)<br/>or micro-batching | [RDB Loader](/docs/pipeline-components-and-applications/loaders-storage-targets/snowplow-rdb-loader/index.md) | Production-ready |
+| BigQuery | Streaming | [BigQuery Loader](/docs/pipeline-components-and-applications/loaders-storage-targets/bigquery-loader/index.md) | Production-ready |
```
stanch, line 28: I think it’s “Early release”?

stanch (approved):

stanch:
This will now need a rebase due to #915

mscwilson:
Rebased! please check I didn't break anything

docs/destinations/warehouses-lakes/querying-data/index.md
```diff
@@ -80,12 +80,16 @@ You can query fields in the self-describing event like so:
...
     ...
-    unstruct_event_my_example_event_1_0_0.my_field,
+    unstruct_event_my_example_event_1.my_field,
     ...
 FROM
     <events>
 ```
 
+:::note
+The column name produced by previous versions of the BigQuery Loader (<2.0.0) would contain full schema version, e.g. `unstruct_event_my_example_event_1_0_0`
```
istreeter, line 90: @stanch can I get your opinion -- does this note need amending, now that the v2 loader is also capable of loading the legacy-style columns?  Should we separate out the three different scenarios?

- V2 loader in normal mode
- V2 loader in legacy mode
- V1 loader

....or alternatively hide/ignore the detail that V2 loader has a legacy mode.

stanch: I think here we should state the V2 normal behavior. And mention that there is a compatibility mode fully described in the migration guide (linking to it).

PR #883, opened 2024-05-10
Correct Azure permissions for Lighthouse

Andy-Hay:
Remove the permissions added for Lighthouse as they don't appear to do what was expected (as per documentation).  Will look into what we need to replace with.


ianarsenault (approved):
LGTM!

PR #886, opened 2024-05-10
Add action to prep a dbt release given package and version

rlh1994:
https://snplow.atlassian.net/browse/PE-6303

This adds an action we can manually trigger to prepare for a dbt version update - it copies the schema and converts it to the new version, and also updates the version in the components list. This means for releases with no other changes (e.g. bug fixes) we can just click this and then it's good to go.

Note it assigns the person who ran the action as a reviewer, but someone else should still be asked to review anyway.

### DIFF
--- a/.github/scripts/copy_schema.sh
+++ b/.github/scripts/copy_schema.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+package_name="${1}"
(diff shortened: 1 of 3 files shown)


agnessnowplow (approved):
lgtm

PR #888, opened 2024-05-13
Azure permissions for Lighthouse added detail

Andy-Hay:
Add image and link to justification for using the third condition option when creating the IAM role.


ianarsenault (approved):
LGTM!

PR #941, opened 2024-06-21
snowplow-attribution: document cv_tstamp_date requirement on Databricks

jethron:
Following the docs for custom `snowplow__conversions_source` on Databricks means dbt-snowplow-attribution fails when running incrementally because the table will be missing the `cv_tstamp_date` field that is added specifically on databricks/spark targets in dbt-snowplow-unified.

This change mentions the column so users can add it accordingly.

See Zendesk ticket 41279.


agnessnowplow (approved):
LGTM, thanks for the corrections!

PR #954, opened 2024-07-05
Patch Snowbridge testing guide

colmsnowplow:


docs/destinations/forwarding-events/snowbridge/testing/index.md
```diff
@@ -30,13 +30,13 @@ Point some test environment tracking to `localhost:9090`, and your events should
 You can run Snowbridge locally via Docker:
 
 <CodeBlock language="bash">{
-`docker run snowplow/snowbridge:${versions.snowbridge}`
+`docker run --env SNOWBRIDGE_CONFIG_FILE="" --env ACCEPT_LIMITED_USE_LICENSE=yes snowplow/snowbridge:${versions.snowbridge}`
```
istreeter, line 33: So this is needed because the default value of `SNOWBRIDGE_CONFIG_FILE` is `/tmp/config.hcl`?  I quite like that there is a sensible default.  But then it's annoying you need to override the sensible default for these examples.

Did you consider amending your Dockerfile to put a dummy minimal hcl file at `/tmp/config.hcl`?  So then in these examples it would be possible to call `docker run ...` without overriding the `SNOWBRIDGE_CONFIG_FILE` environment variable.  But in production it is still possible to mount a custom hcl file to `/tmp/config.hcl`.

istreeter (approved):

docs/destinations/forwarding-events/snowbridge/testing/index.md
colmsnowplow: There's a [ticket](https://snplow.atlassian.net/browse/PDP-1303) to fix the behaviour in Snowbridge, this PR just doesn't want to wait for that release.

I did think of that as an option but unsure if it works as [a quick google](https://superuser.com/questions/1530595/can-we-mount-a-folder-into-docker-such-that-it-overwrites-the-sub-directories-if) makes me suspect that we might then not be able to easily overwrite it when mounting the config file (needs testing at least).

There's an argument that the behaviour belongs in the app too, but I'm not really too bothered about it in practice

PR #958, opened 2024-07-11
Prepare for MDX v3 migration

stanch:
Docusaurus v3 brings MDX v3, which has some breaking changes. The [upgrade guide](https://docusaurus.io/docs/3.0.1/migration/v3#upgrading-mdx) recommends running `npx docusaurus-mdx-checker` while still on Docusaurus v2 / MDX v1 to identify anything that would break.

I’ve run that tool and fixed ~what I could~ all errors. Most whitespace changes are courtesy of VS Code, so I recommend viewing the diff with whitespace changes ignored.

A few things that no longer work in MDX v3:
* `{` / `}` now mean interpolation (even in Markdown mode) and using them for other purposes yields an error
* `<` / `>` can only be used for tags
* Some tags, when followed by text content, expect the closing tag on the same line (the solution is to move the text to the next line)
* Import statements are now fussier and want a blank line afterwards
* `</TabItem>` after a markdown list has always been a problem, but our previous solution of two empty lines no longer works. However, one empty line with no indent works for both MDX v1 and MDX v3!
* Markdown inside `{` `}` is processed as plain text, nor markdown (as per https://mdxjs.com/migrating/v2/#expressions), this is relevant for a couple of files, and the solution is to wrap the markdown in `<ReactMarkdown>`


stanch:
A few things that no longer work in MDX v3:
* `{` / `}` now mean interpolation (even in Markdown mode) and using them for other purposes yields an error
* `<` / `>` can only be used for tags
* Some tags, when followed by text content, expect the closing tag on the same line (the solution is to move the text to the next line)

istreeter (approved):
I skim-read the PR and it all looks reasonable.  And if you say it works, then I'm sure you're right.

Happy to approve this.

Jack-Keene (approved):
lgtm!

PR #960, opened 2024-07-11
Upgrade to Docusaurus 3.4

stanch:


stanch:
Note that some changes are from #958. I did not want to base on that branch because then the preview would not trigger.

PR #977, opened 2024-08-13
Add anonymous id nullification example

agnessnowplow:
Related to: https://snplow.atlassian.net/browse/PE-6563

It far from ideal that we have to handle it like this but this is what we have time for now. People have been complaining that the user_sql field is hard to understand so adding an example from that point of view is a bonus at least.

### DIFF
--- a/docs/modeling-your-data/modeling-your-data-with-dbt/package-features/custom-identifiers/index.md
+++ b/docs/modeling-your-data/modeling-your-data-with-dbt/package-features/custom-identifiers/index.md
@@ -254,7 +254,7 @@ Instead, suppose you want to take the `DATE` value of your `derived_tstamp` as y
  ```yml title="dbt_project.yml"
 vars:
     ...
-    snowplow__session_sql: "e.domain_userid || '_' || e.domain_sessionid"
+    snowplow__session_sql: "DATE(e.derived_tstamp)"
     ...
 ```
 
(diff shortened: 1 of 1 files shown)


docs/modeling-your-data/modeling-your-data-with-dbt/package-features/custom-identifiers/index.md
```diff
@@ -269,4 +268,22 @@ SELECT
...
+Customizing user identifiers works in the exact same way as customizing session identifiers, although you need to make use of the `snowplow__user_identifiers` variable instead of the `snowplow__session_identifiers`, and `snowplow__user_sql` in place of `snowplow__session_sql`. By default the user identifier is the `domain_userid` field which is found in the atomic events table. 
+
+You can find a project with examples to demonstrate this [here](https://github.com/snowplow-incubator/dbt-example-project/tree/main/custom_users).
+
+### Handling Anonymized Users
+In case of applying Client-side anonymisation with session tracking, the `userId` property of the `contexts_com_snowplowanalytics_snowplow_client_session_1` equates to a null UUID which will appear as `00000000-0000-0000-0000-000000000000` in the database. It may be convenient to make this field an actual NULL field to make it easier to exlude them from modeling (e.g the user mapping table of the Unified Package excludes null values). This can be made possible with the use of the `snowplow__user_sql` variable, however, this means that the extraction from the relevant context/sde field needs to be handled manually.
```
matus-tomlein, line 276:
```suggestion
In case of applying Client-side anonymisation with session tracking, the `userId` property of the `contexts_com_snowplowanalytics_snowplow_client_session_1` equates to a null UUID which will appear as `00000000-0000-0000-0000-000000000000` in the database. It may be convenient to make this field an actual NULL field to make it easier to exclude them from modeling (e.g the user mapping table of the Unified Package excludes null values). This can be made possible with the use of the `snowplow__user_sql` variable, however, this means that the extraction from the relevant context/sde field needs to be handled manually. 
```

matus-tomlein (approved):
LGTM!

PR #983, opened 2024-08-19
Add tracked React File Player

greg-el:
This PR adds the `TrackedReactFilePlayer` component, a wrapped version of the `FilePlayer` from `react-player`. This should function as a drop-in replacement to enable tracking for videos currently on the docs.

A few other minor changes to deps:
- The snowplow browser tracker versions have been adjusted slightly, as it was having trouble trying to add the media package whilst on `^3.5.0` (should we be pinning to a specific version anyway?)
- `@docusaurus/types` and `typescript` have been added, as the component is written in typescript

### DIFF
--- a/src/components/TrackedReactFilePlayer/index.tsx
+++ b/src/components/TrackedReactFilePlayer/index.tsx
@@ -1,0 +1,131 @@
+import React, { useRef, useEffect, RefObject } from 'react'
+import FilePlayer, { FilePlayerProps } from 'react-player/file'
+import { v4 as uuidv4 } from 'uuid'
+
+import {
+  trackMediaReady,
+  trackMediaPlay,
+  trackMediaPause,
+  trackMediaBufferStart,
+  trackMediaBufferEnd,
+  trackMediaEnd,
+  trackMediaError,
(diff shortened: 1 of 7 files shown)


src/components/TrackedReactFilePlayer/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+  // Ensure that an `id` prop is provided to prevent bad events
+  if (!id) {
+    throw new Error('TrackedFilePlayer requires an `id` prop')
+  }
+
+  const playerRef = useRef<FilePlayer>(null)
+
+  useEffect(() => {
+    startMediaTracking({ id })
```
matus-tomlein, line 72: The `id` of the media tracking needs to be unique for each viewing. If I understand correctly, here it is the same for each player. Ideally this could generate a new UUID each time the player is initialized.

greg-el: Makes sense, have updated to use a UUID instead.

matus-tomlein (approved):
LGTM!

PR #986, opened 2024-08-22
Add instructions for using the .NET tracker on MAUI apps

matus-tomlein:
Adds a section in the .NET tracker setup guide to explain how to set up the tracker on MAUI apps. The difference is that one can't use our PlatformExtensions package, but the docs now provide a simple snippet for one function that is necessary to initialize the tracker.


mscwilson (approved):
LGTM!

PR #996, opened 2024-08-30
Add links to plugins used by GTM Tag Template

jethron:
Especially for the options that now have multiple plugins (Consent, E-commerce), this disambiguates what each setting actually configures.


matus-tomlein (approved):
LGTM!

PR #999, opened 2024-09-05
Add Snowbridge `snowplow_mode` details

jethron:
Per snowplow/snowbridge#365


colmsnowplow (approved):
Wow this is a brilliant update, thanks so much @jethron!

PR #1001, opened 2024-09-09
Snowbridge 2.4.2

colmsnowplow:


docs/destinations/forwarding-events/snowbridge/concepts/failure-model/index.md
```diff
@@ -22,6 +22,8 @@ This is where a request to the destination technology fails or is rejected - for
 
 Note that this means failures on the receiving end (eg. if an endpoint is unavailable), mean Snowbridge will continue to attempt to process the data until the issue is fixed.
 
+As of Snowbridge 2.4.2, the kinesis target does not treat kinesis throttling as this type of failure. Rather it has an in-built backoff and retry, which will persist until each event in the batch is either successful, or fails for a different reason.
```
istreeter, line 25: I paused on the word "throttling" to consider if readers of these docs would know what it means.  I searched google for kinesis + throttling.  There are hits in the AWS docs for read throttling but not for write throttling.  A longer alternative would be "throttled for exceeding a shard's write throughput".

I don't know if that's needed or worse, but thought I'd mention.

istreeter (approved):
Happy to approve anyway, even though I left a comment.

docs/destinations/forwarding-events/snowbridge/concepts/failure-model/index.md
colmsnowplow: Very good point! Pushed a commit to simply say 'write throughput exceptions' - which should be much clearer. Is that better?

istreeter (approved):
Better :)

PR #1011, opened 2024-09-19
Update tracker versions

mscwilson:
👀


johnmicahreid:
What does this PR do exactly? Is it just showing the latest versions of each tracker?

mscwilson:
this file is supposed to be updated when a new tracker/component version is released, but we've missed a few

johnmicahreid (approved):
LGTM

PR #1019, opened 2024-10-02
Removed docs that dictate that Iglu Server is required by some components

jbeemster:
Also removed some weird characters `<0xa0>` from the docs and replaced them with normal whitespace.

---

Ultimately doesn't change the quickstart flow stating that users should setup an Iglu Server but also isn't misleading that its required (as it isn't).

Made a note that if using the static registry for everything that you shouldn't skip schema lists as well given some apps do require that still.

### DIFF
--- a/docs/pipeline-components-and-applications/iglu/igluctl-2/index.md
+++ b/docs/pipeline-components-and-applications/iglu/igluctl-2/index.md
@@ -16,19 +16,19 @@ Iglu is a schema repository for JSON Schema. A schema repository (sometimes call
 
 Iglu provides a CLI application, called igluctl which allows you to perform most common tasks on Iglu registry. So far, the overall structure of igluctl commands looks like the following:
 
-- `lint` - validate set of JSON Schemas for syntax and consistency of their properties
-- `static` - work with static Iglu registry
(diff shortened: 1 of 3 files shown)


docs/pipeline-components-and-applications/iglu/igluctl-2/index.md
```diff
@@ -172,58 +172,58 @@ One of the more problematic scenarios to handle when generating Redshift table d
...
-- `region` - AWS S3 region. Default: `us-west-2`
-- `skip-schema-lists` - Do not generate and upload schema list objects.
+- `input` - path to your files. Required.
+- `bucket` - S3 bucket name. Required.
+- `s3path` - optional S3 path to prepend your input root. Usually you don’t need it.
+- `accessKeyId` - your AWS Access Key Id. This may or or may not be required, depending on your preferred authentication option.
+- `secretAccessKey` - your AWS Secret Access Key. This may or or may not be required, depending on your preferred authentication option.
+- `profile` - your AWS profile name. This may or or may not be required, depending on your preferred authentication option.
+- `region` - AWS S3 region. Default: `us-west-2`
+- `skip-schema-lists` - Do not generate and upload schema list objects. If using a static registry for all Snowplow applications, don’t enable this setting as some components still require lists to function correctly.
```
jbeemster, line 216: Added the note here @stanch

stanch (approved):

PR #1053, opened 2024-10-25
Update index.md

danielyyy4:
Hi team - I would like to change the "Caution" section from "Caution" to "Info" or "Note" OR ca we remove the note that it is in public preview or is this necessary? @johnmicahreid


danielyyy4:
I agree with removing @johnmicahreid

danielyyy4:
I understand, @johnmicahreid. I currently do not have permissions to merge, but when Nick comes back I will ask him to add me to the Snowplow group so that I can

PR #1060, opened 2024-10-31
Fixing links and updating page to refer to V4

jborlase-snowplow:
Adding reference and links to the JavaScript V4 tracker. 

Might be worth simply referring to the latest version in future to make this page less likely to go out of date.


gibbok-snowplow (approved):
LGTM

PR #1062, opened 2024-11-04
Add Base Data Products tracking tutorial

igneel64:


greg-el (commented):
It seems it doesn't play nice with `index.md` existing, Docusaurus wants to treat it as the home page for `/data-products-base-tracking`, which is breaking the pagination/steps.

I'll have a think about how to handle that, but for now could you rename it to something that isn't `index.md` 😄.

gibbok-snowplow:
@igneel64 just a minor comment from me, please consider renaming files from:

`create-dp.png` to something like `create-data-product.png` as it can help for SEO

igneel64:
@greg-el Changed it to `start.md` 😆

gibbok-snowplow (approved):

PR #1064, opened 2024-11-04
Updating HTML5 Media Tracking Documentation

jborlase-snowplow:
- Fixing error with example (ended event is incorrect)
- Updating example to show filterOutRepeatedEvents usage
- Updating documentation with filterOutRepeatedEvents description and options which was previously not mentioned.

### DIFF
--- a/docs/collecting-data/collecting-from-own-applications/javascript-trackers/web-tracker/tracking-events/media/html5/index.md
+++ b/docs/collecting-data/collecting-from-own-applications/javascript-trackers/web-tracker/tracking-events/media/html5/index.md
@@ -154,6 +154,16 @@ This function begins tracking media events for a given media element. It takes a
 
 ---
 
+#### Filter Out Repeated Events Configuration (Optional)
+
+| Parameter | Type | Description | Default Value |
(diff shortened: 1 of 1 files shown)


docs/collecting-data/collecting-from-own-applications/javascript-trackers/web-tracker/tracking-events/media/html5/index.md
```diff
@@ -154,21 +154,35 @@ This function begins tracking media events for a given media element. It takes a
 
 ---
 
+#### Filter Out Repeated Events Configuration (Optional)
+
+| Parameter | Type | Description | Default Value |
+|-----------|------|-------------|---------------|
```
matus-tomlein, line 160: The `config.filterOutRepeatedEvents` parameter also accepts a boolean which can be used to enable/disable all repeated events regardless of the type, maybe worth adding a mention of that?

```
| `config.filterOutRepeatedEvents` | `boolean` | Whether to filter out seek start, seek end and volume change events. | `false` |
```

matus-tomlein (approved):
LGTM! Thanks for improving the docs for this

PR #1071, opened 2024-11-08
Add performance navigation timing option to full `newTracker` example

greg-el:


docs/collecting-data/collecting-from-own-applications/javascript-trackers/web-tracker/tracker-setup/initialization-options/index.md
```diff
@@ -137,7 +137,8 @@ snowplow('newTracker', 'sp', '{{collector_url_here}}', {
     webPage: true, // Default
     session: false, // Adds client session context entity to events, off by default. Available in v3.5+.
     browser: false, // Adds browser context entity to events, off by default. Available in v3.9+.
-    performanceTiming: true,
+    performanceTiming: false, // Superseded by `performanceNavigationTiming` in v4
+    performanceNavigationTiming: true, // Introduced in v4
```
matus-tomlein, line 141: Is this option there? Seems like we only have the `performanceTiming` one: https://github.com/snowplow/snowplow-javascript-tracker/blob/master/trackers/javascript-tracker/src/configuration.ts#L37

So currently I think it's not possible to disable the performance navigation timing context, which is not ideal... We should add an option to do that.

greg-el: Oh do'h, I was reading the performance navigation PR and thought the option was propagated through like `performanceTiming`. I'll leave this open for now until we've done so 😄.

PR #1073, opened 2024-11-12
Add documentation for what information is used to calculate the media session entity

matus-tomlein:


matus-tomlein:
Very good point! Have added a table like that to show the relation with the media player properties and evenets.

Jack-Keene (approved):
LGTM!

PR #1074, opened 2024-11-12
Update passthrough instructions

agnessnowplow:
We need to update the passthrough field extraction instructions as more and more users find it ambiguous which leads to problems.

### DIFF
--- a/docs/modeling-your-data/modeling-your-data-with-dbt/package-features/passthrough-fields/index.md
+++ b/docs/modeling-your-data/modeling-your-data-with-dbt/package-features/passthrough-fields/index.md
@@ -20,110 +20,105 @@ Passthrough fields are the term used to define any field from your [events this 
 
 ## Usage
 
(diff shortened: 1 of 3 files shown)


docs/modeling-your-data/modeling-your-data-with-dbt/package-features/passthrough-fields/index.md
```diff
@@ -26,15 +30,148 @@ vars:
     snowplow__view_passthroughs: ['v_collector', {'sql': 'event_id || app_id', 'alias': 'event_app_id'}]
 ```
 
-would add the `v_collector` field and a field called `event_app_id` which is the concatenation of the `event_id` and the `app_id`. A more useful case for the SQL block is to extract a specific field from a context, however you can also just use the context column name as the field to bring the whole context through as is e.g.
+would add the `v_collector` field and a field called `event_app_id` which is the concatenation of the `event_id` and the `app_id`. A more useful case for the SQL block is to extract a specific field from an entity, however you can also just use the context column name as the field to bring the whole column through as is e.g.
```
ilias1111, line 33: We change the word context to entity ( makes sense I guess ) but we still have the 
"context column name". How thorough we should be with renaming context to entity when we cannot change it in the code / package for breaking reasons

agnessnowplow: Yeah, that's why I opted for the blended wording. I think our modeling docs still mainly reflect the old context naming as we are reliant on the fields. However, to make it clear for users what we are talking about, whenever we see only context in one place we should add the sde/entity naming at least once so that they can link it to the rest of the product, I think.

ilias1111 (approved):

PR #1076, opened 2024-11-12
Add v4, remove v2 from third party CDN hosting

greg-el:
This PR removes mentions of JavaScript v2 from the third party CDN hosts, and adds in v4.


jethron (approved):
LGTM!

PR #1077, opened 2024-11-13
Add a clarification to the media session stats calculation about ad playback affecting time played metrics

matus-tomlein:


Jack-Keene (approved):
LGTM

PR #1082, opened 2024-11-18
Restructure the sidebar

mscwilson:


johnmicahreid:
> should the recipes move somewhere as part of this PR?
Can you just put them in Resources for now? Eventually they will be deprecated. 

> in API Reference, why is it Stream Collector and not just Collector? Historical reasons?
No idea, maybe @stanch knows

> are we doing Title or Sentence case for the sidebar (and page titles)?
I think we should do Title Case.

johnmicahreid:
Overall: 

- Looks good, it's much cleaner! 
- I think the sidebar now needs some love (this could be done in a separate task) e.g. icons, padding, font, toggle behavior. I'll get some help from Design on this one. 
- The empty pages with just a list of child items are kinda bothering me now - it makes the experience feel more 'clicky'. A quick win might be to hide the pages where this happens, so you have to click into the next item on the sidebar to see them. @stanch I believe you said this is possible? 
- Can you remove the `Data Products` folder under `Data Product Studio` and nest the child items at the same level as `Data Quality`? It just feels like it's too many folders

johnmicahreid (approved):
Added comments separately, LGTM

stanch (approved):

PR #1083, opened 2024-11-21
Update JS tracker in docs to version 4.0.4

matus-tomlein:
Updates the JavaScript tracker dependency to the latest version 4.0.4. This makes it no longer necessary to call `refreshLinkClickTracking`, so I was able to remove a bunch of code handling that.


jethron (approved):
Less code is best code, LGTM!

PR #1111, opened 2024-12-31
Add more VM SKUs for Azure

Andy-Hay:
- Update https://docs.snowplow.io/docs/get-started/snowplow-bdp/private-managed-cloud/setup-guide-azure/#check-subscription-resource-availability to add additional SKUs
- Wrap the text on some sections so that its more easily readable when editing
- Add in more detail about what to do if SKUs aren't available in all three availability zones in subscription


antonis-snowplow (approved):
LGTM

PR #1121, opened 2025-01-09
snowplow-cli: promote schema use for data products and source apps

jamessnowplow:
I think its enough to point people at it and not try to shoe horn an intro to lsp use in there.

### DIFF
--- a/docs/resources/recipes-tutorials/recipe-data-structures-in-git/index.md
+++ b/docs/resources/recipes-tutorials/recipe-data-structures-in-git/index.md
@@ -401,6 +401,10 @@ data:
 * `resourceName` is a unique identifier of the source applications. It must be a valid uuid v4
 * `data` is the contents of the source app
 
+:::note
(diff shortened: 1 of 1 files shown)


cksnp (approved):

docs/resources/recipes-tutorials/recipe-data-structures-in-git/index.md
```diff
@@ -475,7 +483,7 @@ data:
 ```
 
 :::note
-You'll need to come up with a valid uuid V4 for the `resourceName` of an event specification. You can do so by using an (online generator)[https://www.uuidgenerator.net], or running the `uuidgen` command in your terminal
+You'll need to come up with a valid uuid V4 for the `resourceName` of an event specification. You can do so by using an [online generator](https://www.uuidgenerator.net), or running the `uuidgen` command in your terminal
```
gleb-lobov, line 486: 🫠

jamessnowplow: I only noticed because I did it (that)[way around] and got confused when the links didn't work.

PR #1123, opened 2025-01-16
Clarify MFA for SnowplowDeployment role

Andy-Hay:
Clarify how SnowplowDeployment needs `Require MFA` to be set, so that we do not have it set to false in IAM condition as it means that we can't connect to it to perform any investigations


stdfalse (approved):
LGTM!

ianarsenault (approved):

PR #1130, opened 2025-01-29
Add Snowplow style guide

mscwilson:


src/pages/style-guide.md
```diff
@@ -0,0 +1,556 @@
...
+  | JSON schemas | JSONs |
+  | APIs         |       |
+  | SDKs         |       |
+  | URLs         |       |
+
+- For the **documentation** website, we have a docs plugin that adds a dotted line and a tooltip explanation to acronyms, e.g. hovering over "BDP" will show that it stands for "Behavioral Data Platform": add new acronym definitions to the `src/remark/abbreviations.js` file to enable this behavior
+- (Technically, these are all initialisms, not acronyms)
+
+### Capitalization
+- Third-party products should be capitalized as appropriate
```
colmsnowplow, line 284: Minor suggestion - not just third-party products, ours too!

Might we add Enrich to this list - which comes with the benefit of clarifying a small piece of nuance in the next example (enrichment is a process that shouldn't be capitalised, Enrich is a product that should - if I have it right).

IMO there's nothing wrong at all with how it's written as-is, I just thought I'd share the idea :)

colmsnowplow: Oh, I continued reading and there's a snowplow specific bit. So perhaps this comment is moot - apologies for the noise!

mscwilson: I'll move the Snowplow bit up so it's more obvious

src/pages/style-guide.md
```diff
@@ -0,0 +1,556 @@
...
+- Use inline code (single backticks, <code>`</code>) for field names, classes, file names, etc.
+
+    | ✅                             |
+    | ----------------------------- |
+    | a single `geo_location` field |
+    | a `event_id` UUID string      |
+    | tracking a `ScreenView` event |
+
+- Use code blocks (triple backticks, <code>```</code>) for code examples
+  - Specify the language next to the opening backticks, so that the code block is rendered correctly
```
colmsnowplow, line 552: If this doesn't render appropriately, the language might need to be added to [this config](https://github.com/snowplow/documentation/blob/cfa199351c1e2b86ff1c70d65bdf8a7009a8d976/docusaurus.config.js#L157-L178)

PR #1131, opened 2025-01-30
Rename data apps to visualizations

mscwilson:
Please have a quick check that I didn't change it to say something incorrect, especially on the overview page.


johnmicahreid:
I know the task was literally to rename Data Applications to Data Model Packs, but I think there's a distinction between the SKU (what we are packaging and selling to customers i.e. the Data Model Pack) and the actual products. It doesn't make sense to call the apps data model packs because ...they aren't - especially not apps like the Funnel Builder. 

Discussing with @danielyyy4 (Daniela) on the Console renaming, we agreed to call them Visualizations (in Console). Could you update the terminology in this PR accordingly?

agnessnowplow (approved):
Makes sense to me, love the care that went into this one!

PR #1133, opened 2025-01-31
Added snowtype java example

gleb-lobov:


docs/data-product-studio/snowtype/using-the-cli/index.md
johnmicahreid, line 1: It looks a little weird to see those SelfDescribing calls in there - I thought the point of Snowtype was that it would generate function calls? But maybe I just don't understand Java syntax

johnmicahreid (approved):
LGTM overall

docs/data-product-studio/snowtype/using-the-cli/index.md
gleb-lobov: In my understanding it looks funny because Java is being Java and we are only using data structures, not event specs. For event specs we generate a function that returns a `SelfDescribing`, so the problem is not present. But for data structures we don't, so we have to build it. Other languages look better, because they have "better" tracker API. 

@igneel64 could you please confirm that this looks ok to you and I didn't miss something?

igneel64: Looking at it from a consumer perspective I would love to have a `.toEvent()` and `.toEntity()` method as Kotlin and Swift have for Snowtype cases. Not sure if this is doable based on the tracker API and if it is, it would need a bit more extra work now. Depends on what we want.

For Event Specifications, there are automatically generated methods as Gleb said so we are ok there.

gleb-lobov: There is no `Event` or `Entity` types in Java tracker API. There is a `SelfDescibinJson` that represents both events and entities and `SelfDescribing` that represents an event with optional array of entities. So `SelfDescribing` contains one `SelfDescibinJson` and an optional array of `SelfDescibinJson`. To replicate `.toEvent()` and `.toEntity()` I've added the `.toSelfDescribingJson()`

mscwilson:
what's left to do on this PR? is it ready to merge?

gleb-lobov:
@mscwilson it's clear to be merged from my side, I think we clarified all the concerns

PR #1143, opened 2025-02-10
Use environment variables

stanch:


mscwilson (approved):

stanch:
Does not seem to work, will need to try something else

PR #1153, opened 2025-02-24
Fix references to the renamed snowplow-javascript-tracker-examples repo

matus-tomlein:


mscwilson:
these are in the process of being fixed I think in #1149

PR #1177, opened 2025-03-24
Update paths in Kafka Live Viewer Profiles accelerator

mscwilson:
The repo that the accelerator is based on was updated


trentusus (approved):
looks good

PR #1181, opened 2025-03-27
Add Product Fruits

mscwilson:


src/theme/Root.js
```diff
@@ -1,11 +1,47 @@
...
 import theme from '@site/src/components/MuiTheme'
+import { ProductFruits } from 'react-product-fruits'
+import useDocusaurusContext from '@docusaurus/useDocusaurusContext'
+
+const useCookie = () => {
+  const [userIdValue, setUserIdValue] = useState('')
+  const [sessionIdValue, setSessionIdValue] = useState('')
+
+  useEffect(() => {
+    const cookie = document.cookie
```
gibbok-snowplow, line 13: should we use decodeURIComponent when reading cookies?

src/theme/Root.js
```diff
@@ -1,11 +1,47 @@
...
+      .split('; ')
+      .find((row) => row.startsWith('_sp_biz1_id'))
+
+    const cookieExists = cookie !== undefined && cookie !== null
+
+    setUserIdValue(cookieExists ? cookie.split('=')[1].split('.')[0] : '')
+    setSessionIdValue(cookieExists ? cookie.split('=')[1].split('.')[5] : '')
+  })
```
gibbok-snowplow, line 21: do we really need to run the useEffect on every re-render of the component? if not, please include an empty array as dependencies of useEffect so the reading of cookies happens only on mount (once)

gibbok-snowplow: maybe we can remove some repetitions in the split part for instance if you like it:
Just some pseudo code here (please double check it 😄 )

```
const useCookie = () => {
  const [userId, setUserId] = useState('unknown_user') // add the default state
  const [sessionId, setSessionId] = useState('unknown_session') // same here

  useEffect(() => {
    const cookieValue = document.cookie
      .split('; ')
      .find((row) => row.startsWith('_sp_biz1_id='))
      ?.split('=')[1] // add split here

    if (cookieValue) {
      const parts = cookieValue.split('.')
      setUserId(parts[0] ?? 'unknown_user')
      setSessionId(parts[5] ?? 'unknown_session')
    }
  }, []) // only run on mount instead of every render

  return [userId, sessionId]
}

```

gibbok-snowplow:
are you really sure the package-lock.json file must be included? this is a yarn project only, if I recall correctly only yarn lock file should be necessary

docusaurus.config.js
```diff
@@ -26,6 +26,10 @@ module.exports = {
     defaultLocale: 'en',
     locales: ['en'],
   },
+  customFields: {
+    productFruitsNext: process.env.PRODUCT_FRUITS_NEXT,
+    productFruitsProd: process.env.PRODUCT_FRUITS_PROD,
```
gibbok-snowplow, line 31: may I ask, why 2 properties next and prod?
I thought would be only a single property, then you set in the .env the right info for next or prod

customFields {productFruits:'xxxx'}

src/theme/Root.js
```diff
@@ -1,11 +1,47 @@
...
+    username: userId,
+    props: { sessionId: sessionId },
+  }
   return (
     <>
+      <ProductFruits
+        workspaceCode={siteConfig.customFields.productFruitsNext}
+        language="en"
+        user={userInfo}
+        lifeCycle="unmount"
```
gibbok-snowplow, line 42: are you sure for lifecycle unmount?

https://help.productfruits.com/en/article/developers-integration-react-nextjs#2.-find-the-right-place-where-to-put-the-component

gibbok-snowplow:
@mscwilson Thanks for the PR, I just left few comments, if you need ping me any time. Thanks

src/theme/Root.js
mscwilson: I wasn't sure where/when I'd call the `destroy()` function otherwise - any suggestions?

src/theme/Root.js
mscwilson: what is this for?

gibbok-snowplow: This ensures special characters are properly interpreted.

src/theme/Root.js
gibbok-snowplow: Did you fully tested, have you noticed any flickering if not, I think you can leave it

gibbok-snowplow:
I think you need to remove the package-lock.json from the pr

gibbok-snowplow (approved):
LGTM

PR #1183, opened 2025-03-28
Single scoped azure custom role

Andy-Hay:
Based on feedback from a customer issue, include that the custom role should be single scoped.


ianarsenault (approved):
LGTM!

PR #1190, opened 2025-04-04
Update for JS tracker 4.5 and iOS tracker 6.2.1 releases

matus-tomlein:
Update versions and add docs for JS tracker changes


mscwilson (approved):

src/componentVersions.js
```diff
@@ -6,9 +6,9 @@ export const versions = {
   flutterTracker: '0.7.1',
   golangTracker: '3.1.0',
   googleAmpTracker: '1.1.0',
-  iosTracker: '6.2.0',
+  iosTracker: '6.2.1',
   javaTracker: '2.1.0',
-  javaScriptTracker: '4.3.0',
+  javaScriptTracker: '4.4.1',
```
jethron, line 11: Late to the party sorry, but shouldn't this be 4.5.0 rather than 4.4.1? Not sure [we have a 4.4.1](https://github.com/snowplow/snowplow-javascript-tracker/releases)?

Issue #1198, opened 2025-04-08
Inconsistent advice querying lake loader/Synapse data


jethron:
In the "Synapse Analytics" tab at [How schema definitions translate to the warehouse](https://docs.snowplow.io/docs/destinations/warehouses-lakes/schemas-in-warehouse/#location) (which doesn't use the `warehouse=synapse` parameter for some reason), it says:

> The name of each JSON field is the name of the schema field converted to snake case.

And gives an example of a `lastName` schema property being queried as `last_name`.

But in [Querying Snowplow data](https://docs.snowplow.io/docs/destinations/warehouses-lakes/querying-data/?warehouse=synapse#self-describing-events) it gives an example of querying a camel-case field name:

> JSON_VALUE(unstruct_event_my_example_event_1, '$.myField')

From the first page, I would expect this to need to be `JSON_VALUE(unstruct_event_my_example_event_1, '$.my_field')`, but it's not clear which of the two pages is correct.

PR #1201, opened 2025-04-08
Fix field name casing for Synapse

mscwilson:
For issue #1198 

This PR fixes the example that showed a field name in camel case.
Also, it rearranges the tabs on the **Schemas in the warehouse** page, so they're in the same order as in the **Querying data** page.


istreeter (approved):
It's hard to review this, because so much appears to be different on first glance.  But I guess it's all really the same, just in a different order.

In that case I'm happy with it.  From what I can tell, it is all correct.

PR #1202, opened 2025-04-08
Add Extensions to features comparison

mscwilson:


stanch:
I would prefer to keep some of the initialisms as is, as they are commonly used and should be clear in a comparison table (and we already provide definitions on hover). Specifically, SLA and VPC.

stanch (approved):

PR #1210, opened 2025-04-10
Add more details about cross-domain tracking

mscwilson:
I took bits out of the reusable partial to use directly in the JS tracker page because it was easier to edit and break it up with subheadings


docs/sources/trackers/javascript-trackers/web-tracker/cross-domain-tracking/index.md
```diff
@@ -4,23 +4,351 @@ date: "2022-08-30"
...
+
+:::tip
+
+If you enable link decoration, aim for at least one event to be tracked on the page. The tracker writes the `domain_userid` to a cookie when it tracks an event. If the cookie doesn't exist when the user leaves the page, the tracker will generate a new ID for them when they return, rather than keeping the old ID. This can make user stitching difficult.
+
+:::
+
+## Extended format
+
+The extended `_sp` value always includes `domainUserId` and `timestamp`. The other properties are configurable using the `ExtendedCrossDomainLinkerAttributes` object.
```
jethron, line 51: The `ExtendedCrossDomainLinkerAttributes` type name isn't useful by itself unless we link to e.g. https://snowplow.github.io/snowplow-javascript-tracker/docs/browser-tracker/markdown/browser-tracker.extendedcrossdomainlinkerattributes

Then (pedantically, sorry) given it's the type it'd be "_a_ `ExtendedCrossDomainLinkerAttributes` object", rather than "_the_ `ExtendedCrossDomainLinkerAttributes` object" - the "the" is the `useExtendedCrossDomainLinker` object.

docs/sources/trackers/javascript-trackers/web-tracker/cross-domain-tracking/index.md
```diff
@@ -4,23 +4,351 @@ date: "2022-08-30"
...
+
+By default, `reason` will be the link text from a cross-domain link.
```
jethron, line 53: This is only the default if you specify `true` for `reason` instead of providing your own callback, the default is that it's not collected at all, only `sessionId` and `sourceId` are added with `useExtendedCrossDomainLinker: true`.

docs/sources/trackers/javascript-trackers/web-tracker/cross-domain-tracking/index.md
```diff
@@ -4,23 +4,351 @@ date: "2022-08-30"
...
+<Tabs groupId="platform" queryString>
+  <TabItem value="js" label="JavaScript (tag)" default>
+
+```javascript
+snowplow('newTracker', 'namespace', '{{collector_url_here}}', {
+  appId: 'my-app-id',
+  appVersion: '0.1.0',
+  cookieSameSite: 'Lax',
+  crossDomainLinker: function (linkElement) {
+    return (linkElement.href === 'http://acme.de' || linkElement.id === 'crossDomainLink');
```
jethron, line 145: This doesn't match the `http://acme.de/` link in the prose.

docs/sources/trackers/javascript-trackers/web-tracker/cross-domain-tracking/index.md
```diff
@@ -4,23 +4,351 @@ date: "2022-08-30"
...
+
+```javascript
+{
+  crossDomainLinker: function (linkElement) {
+    return /^https:\/\/github\.com/.test(linkElement.href);
+  }
+}
+```
+
+You can also call the `crossDomainLinker` function directly:
```
jethron, line 177: ~This probably needs an explainer; you would want to do this because it follows the old link tracking behaviour where it only adds the event listener to decorate the link to links that already exist on the page. If you're in an SPA or have some other dynamic content, those links won't be decorated even if they would satisfy your callback.~

~So it's probably something you'd want to do after each `trackPageView` call to make everything up to date.~

Oh, saw the below section, this should just be removed and left to the "Update event listeners" section.

docs/sources/trackers/javascript-trackers/web-tracker/cross-domain-tracking/index.md
```diff
@@ -4,23 +4,351 @@ date: "2022-08-30"
...
+});
+```
 
   </TabItem>
 </Tabs>
+
+
+### Decorate all links
+
+To decorate every link, regardless of its destination:
```
jethron, line 203: We shouldn't suggest doing this, or at least warn against it.

docs/sources/trackers/javascript-trackers/web-tracker/cross-domain-tracking/index.md
```diff
@@ -4,23 +4,351 @@ date: "2022-08-30"
...
+```javascript
+crossDomainLinker(function(linkElement) {
+  return linkElement.href.indexOf('javascript:') < 0;
+});
+```
+
+  </TabItem>
+</Tabs>
+
+Alternatively, only count links that parse as web URLs by checking the link's [`hostname`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLAnchorElement/hostname). This should automatically exclude links that don't lead simply to other web pages.
```
jethron, line 236: > This should automatically exclude links that don't lead simply to other web pages.

I'm not sure what this means?

Probably more useful to check the hostname against an array of domains or something.

docs/sources/trackers/javascript-trackers/web-tracker/cross-domain-tracking/index.md
```diff
@@ -4,23 +4,351 @@ date: "2022-08-30"
...
+}
+```
+
+Link decoration will only occur when links are to different sites within the list.
+
+### Update event listeners
+
+When the tracker loads it doesn't immediately decorate links. Instead, it adds event listeners to links which decorate them when a user clicks on them or navigates to them using the keyboard. This ensures that the timestamp added to the querystring is fresh.
+
+If further links get added to the page after the tracker has loaded, you can use the tracker's `crossDomainLinker` method to add listeners again. Listeners won't be added to links which already have them.
```
jethron, line 294: ... so it's good practice to make sure you use the same linker function every call, it won't be updated.

docs/sources/trackers/javascript-trackers/web-tracker/cross-domain-tracking/index.md
```diff
@@ -4,23 +4,351 @@ date: "2022-08-30"
...
+* Extended format: the querystring can include more parameters, but the tracker will decorate all cross-domain links
+
+If you set both configuration options, the tracker will use the extended format.
+
+You can include the following properties in the `_sp` querystring:
+
+| Property        | Description                                    | Extended | Short |
+| --------------- | ---------------------------------------------- | -------- | ----- |
+| `domainUserId`  | Current tracker-generated UUID user identifier | ✅        | ✅     |
+| `timestamp`     | Current epoch timestamp, ms precision          | ✅        | ✅     |
```
jethron, line 26: Nit: do these columns need width adjustment?

jethron (commented):
Great start, thanks for picking this up!

docs/pipeline/enrichments/available-enrichments/cross-navigation-enrichment/index.md
```diff
@@ -1,22 +1,30 @@
...
+The extended cross-navigation format is `_sp={domainUserId}.{timestamp}.{sessionId}.{subjectUserId}.{sourceId}.{platform}.{reason}`.
+
+If this enrichment isn't enabled, Enrich parses the `_sp` querystring parameter according to the short format `_sp={domainUserId}.{timestamp}`
 
 ## Configuration
 
 - [Schema](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow.enrichments/cross_navigation_config/jsonschema/1-0-0)
 - [Example](https://github.com/snowplow/enrich/blob/master/config/enrichments/cross_navigation_config.json)
 
+```json reference
```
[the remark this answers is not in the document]
jethron: Oh, it's not documented but apparently [we can change](https://github.com/saucelabs/docusaurus-theme-github-codeblock/blob/db0af5832dca98a55d12d83c76174a58201037a9/src/theme/ReferenceCodeBlock/index.tsx#L57) the "See full example on GitHub" text too if that doesn't make sense in this context.

````markdown
```json reference title="Schema" referenceLinkText="See schema on Github"
````

File: docs/pipeline/enrichments/available-enrichments/cross-navigation-enrichment/index.md [hunk 10]
```diff
@@ -28,10 +36,10 @@ import TestingWithMicro from "@site/docs/reusable/test-enrichment-with-micro/_in
 This enrichment extracts `_sp` querystring parameter from the following inputs:
 
 - The `page_url` field from the Snowplow event
-- The referer uri extracted from corresponding HTTP header in the raw event
+- The `referer` URI extracted from corresponding HTTP header in the raw event
 
 ## Output
 
-This enrichment adds a new derived context to the enriched event with [this schema](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/cross_navigation/jsonschema/1-0-0).
+This enrichment adds a new derived entity to the enriched event based on [this schema](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/cross_navigation/jsonschema/1-0-0).
```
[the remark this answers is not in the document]
jethron: We later discovered in #1259 that this is a bad idea for `com.snowplowanalytics.snowplow` vendored schemas because they tend to be adblocked and throw errors on the page when that occurs, so best left as a link..

docs/sources/trackers/mobile-trackers/tracking-events/session-tracking/index.md
```diff
@@ -212,30 +212,23 @@ Snowplow.createTracker(getApplicationContext(), namespace, networkConfig, sessio
...
-This is useful for tracking the movement of users across different apps and platforms.
-It is part of our cross-navigation solution and is equivalent to [cross-domain tracking on the JavaScript tracker](/docs/sources/trackers/web-trackers/cross-domain-tracking/index.md).
-
-For example, calling `decorateLink` on `appSchema://path/to/page` will produce the following result:
+The `decorateLink` function is part of Snowplow's [cross-navigation solution](/docs/events/cross-navigation/index.md) for tracking the movement of users across different apps and platforms.
 
-```
-appSchema://path/to/page?_sp=domainUserId.timestamp.sessionId.subjectUserId.sourceId.platform.reason
-```
+Choose which parameters to include using a `CrossDeviceParameterConfiguration` object. The `domainUserId` and `timestamp` are always included automatically.
```
jethron, line 221: Should this link to the API docs or something? The table below implies to me that we need to pass the values from the Controller/Configurations, but instead we actually just need to pass booleans (or a string for the `reason`) for each field and the tracker collects the values. This is only apparent from the examples further down ATM.

docs/sources/trackers/mobile-trackers/tracking-events/session-tracking/index.md
```diff
@@ -212,30 +212,23 @@ Snowplow.createTracker(getApplicationContext(), namespace, networkConfig, sessio
...
-- `sourceId` – The `appId` (value of `TrackerConfiguration.appId`) – optional.
-- `platform` - The platform of the current device (value of `TrackerController.devicePlatform` – optional.
-- `reason` – Identifier/information for cross-navigation – optional.
+| Property         | Description                      | Value used                         |
+| ---------------- | -------------------------------- | ---------------------------------- |
+| `sessionId`      | Current session UUID identifier  | `SessionController.sessionId`      |
+| `subjectUserId`  | Custom business user identifier  | `SubjectController.userId`         |
+| `sourceId`       | Application identifier           | `TrackerConfiguration.appId`       |
+| `sourcePlatform` | Platform of the current device   | `TrackerController.devicePlatform` |
+| `reason`         | Custom information or identifier | Custom string                      |
```
jethron, line 231: Should this note what is/isn't enabled by default, and perhaps the reasoning for that? E.g. `user_id` isn't enabled by default to prevent accidentally leaking personal data to third party URIs, so requires enabling explicitly, and `platform` can usually be inferred from the source App ID.

docs/sources/trackers/web-trackers/cross-domain-tracking/index.md
```diff
@@ -4,23 +4,239 @@ date: "2022-08-30"
...
 import Tabs from '@theme/Tabs';
 import TabItem from '@theme/TabItem';
-import CrossDomain from "@site/docs/reusable/javascript-tracker-cross-domain/_index.md"
 ```
 
+Configure [cross-domain tracking](/docs/events/cross-navigation/index.md) using the `crossDomainLinker` and `useExtendedCrossDomainLinker` fields of the [configuration object](/docs/sources/trackers/web-trackers/tracker-setup/initialization-options/index.md).
+
+:::tip
+
+If you enable link decoration, aim to track at least one event on the starting page. The tracker writes the `domain_userid` to a cookie when it tracks an event. If the cookie doesn't exist when the user navigates to the cross-domain destination, the tracker will generate a new ID for them when they return, rather than keeping the old ID. This can make user stitching difficult.
```
jethron, line 16: I don't believe this is true anymore? Just creating a tracker (required to set these settings) should be enough to create the cookie in question.

docs/sources/trackers/web-trackers/tracker-setup/initialization-options/index.md
```diff
@@ -53,54 +55,56 @@ newTracker('sp', '{{collector_url_here}}', {
   </TabItem>
 </Tabs>
 
-The tracker will be named `sp` (tracker namespace) and will send events to the a collector url you specify by replacing `{{collector_url_here}}`. The final argument is the configuration object. Here it is just used to set the app ID and the common webPage context for each event. Each event the tracker sends will have an app ID field set to “my-app-id”.
+The tracker will send events to the Collector URL you specify by replacing `{{collector_url_here}}`. The final argument is the configuration object. Here it's used to set the app ID and the webPage entity for each event. Each event the tracker sends will have an `app_id` field set to `my-app-id`.
```
jethron, line 58: This example sets more than just `appId` and `contexts.webPage` now, maybe this needs a revisit?

docs/sources/trackers/web-trackers/cross-domain-tracking/index.md
```diff
@@ -4,23 +4,239 @@ date: "2022-08-30"
...
+  <TabItem value="js" label="JavaScript (tag)" default>
+
+```javascript
+snowplow('newTracker', 'namespace', '{{collector_url_here}}', {
+  appId: 'my-app-id',
+  appVersion: '0.1.0',
+  cookieSameSite: 'Lax',
+  crossDomainLinker: function(defineWhichLinksToDecorate) { },
+  useExtendedCrossDomainLinker: {
+    userId: true;
```
jethron, line 172: Maybe a comment here to state if your `userId` contains personal data to make sure your `crossDomainLinker` won't leak it to third party sites.

docs/events/cross-navigation/index.md
```diff
@@ -0,0 +1,54 @@
...
+title: "Cross-navigation tracking"
+description: "Configure cross-domain tracking to maintain user identity between different domains in your ecosystem"
+date: "2025-07-26"
+sidebar_position: 5
+---
+
+When users navigate between different domains in your ecosystem, their user identity may get fragmented. Examples include navigation from your main website to a partner site, or to a mobile app. This creates gaps in your user journey data and makes it difficult to understand the complete customer experience across your digital properties.
+
+:::info Web subdomains
+This problem doesn't usually apply for navigation between web subdomains, as the web trackers have `discoverRootDomain` [enabled by default](/docs/sources/trackers/web-trackers/tracker-setup/initialization-options/index.md).
```
jethron: Should this have an example to avoid being jargon? 
> (e.g. **www**.example.com and **blog**.example.com)

PR #1224, opened 2025-04-23
Document JS element tracking plugin

jethron:
This adds documentation for the [Element Tracking plugin](https://github.com/snowplow/snowplow-javascript-tracker/pull/1400) for the web trackers.


docs/sources/trackers/javascript-trackers/web-tracker/tracking-events/element-tracking/index.md
```diff
@@ -0,0 +1,793 @@
...
+```
+
+#### Custom context
+
+Custom entities can be attached to the events generated by the plugin.
+
+- You can include `context` alongside `elements` when calling `startElementTracking`. You can pass an array of static entities _or_ a callback function that returns such an array. The function will receive the element that the event is relevant to, and the matching rule that defined the event should fire.
+- Individual rules may also contain specific `context` in the same format as above.
+
+## Examples
```
matus-tomlein, line 402: These examples are super interesting and useful and I think that's what most users would like to see as they are trying to understand what this plugin offers (or want to copy-paste some code to their app). Since they are also wrapped in the dropdown boxes and don't take much space, can we move them to the beginning?

docs/sources/trackers/javascript-trackers/web-tracker/tracking-events/element-tracking/index.md
```diff
@@ -0,0 +1,793 @@
...
+
+<details>
+  <summary>Scroll sections</summary>
+
+  The homepage has content grouped into distinct "layers" as you scroll down the page.
+  To see when users scroll down to each section, we can track an `expose` event for each section.
+  We can capture the header element text to identify each one.
+
+  ```json title="Rule configuration"
+  {
```
matus-tomlein, line 414: I think it would be a bit more user friendly to give the full snippet including the `startElementTracking` JS call? Would take a bit less mental effort to understand and could let people copy stuff more easily.

matus-tomlein (approved):
LGTM!

docs/sources/trackers/javascript-trackers/web-tracker/tracking-events/element-tracking/index.md
jethron: Good call, I tried to avoid this to not have to duplicate a bunch of `Tab` stuff for the different APIs. I'll just add a note mentioning the syntax for the examples.

PR #1233, opened 2025-05-07
Add a top-level events section

mscwilson:
This PR moves the "Snowplow events" section out of Sources>Trackers to the top level of the sidebar.

This will make it somewhere to put info (for now) about concepts used in multiple parts of the product, e.g. cross-domain tracking, global context, or anonymous tracking.

The changes here are just the result of running `./move.sh docs/sources/trackers/snowplow-tracker-protocol docs/snowplow-events`


johnmicahreid:
Can we just call it Events rather than Snowplow Events?

johnmicahreid:
I feel like @stanch should input here - overall I like unnesting some of the Sources content, but it now overlaps a bit with the `fundamentals/events` section

mscwilson:
yeah, I intend to deduplicate that later

johnmicahreid (approved):

PR #1240, opened 2025-05-13
Add note for encodeBase64 default for JS v4 migration guide

jethron:


miike (approved):
Good call with the WAF note, that is important.

PR #1243, opened 2025-05-14
Promote cookie extension service

mscwilson:
Made the info easier to find


johnmicahreid (approved):
Love this, it was super buried and called something confusing before!

PR #1244, opened 2025-05-14
Separate web and Node.js tracker sections

mscwilson:
My quest to reduce sidebar nesting continues


johnmicahreid:
I can't see the deploy preview - also, shall we get @igneel64 to take a look?

igneel64:
I like it :) I would just note somewhere on the introductory pages of the node.js tracker that TypeScript types are also available in the package.

phil-snowplow:
lgtm! One small nit: Can we change the title of **Destinations > Forwarding Events** to "Event Forwarding"? That's what the new Snowbridge-based feature will be called when we release it.

PR #1245, opened 2025-05-14
Update dependencies

mscwilson:
Updating some dependencies in preparation for upgrading to Docusaurus 3. 

The syntax highlighting for GitHub codeblocks is broken (e.g. [here](https://docs.snowplow.io/docs/destinations/forwarding-events/snowbridge/configuration/targets/)), but that's not too terrible


greg-el (approved):
LGTM, can't find anything broken! Looks like a bunch more warnings about peer deps when building, but I figure they'll go away again once Docusaurus is on v3.

PR #1246, opened 2025-05-14
Roku Tracker v0.3.0 changes

jethron:
Update pending https://github.com/snowplow-incubator/snowplow-roku-tracker/pull/19


matus-tomlein (approved):
LGTM, but can you also add Roku tracker to the list of tracking options in the media dbt package now that you have tested it:
https://docs.snowplow.io/docs/modeling-your-data/modeling-your-data-with-dbt/dbt-models/dbt-media-player-data-model/

PR #1247, opened 2025-05-15
Add a page about timestamps

mscwilson:


docs/events/timestamps/index.md
```diff
@@ -0,0 +1,33 @@
...
+sidebar_position: 4
+---
+
+Snowplow events have multiple timestamps that are added as the payload moves through the pipeline.
+
+The timestamps are:
+
+| Timestamp             | Added by                                                                            | Description                                                                                                                                                            | In all events |
+| --------------------- | ----------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------- |
+| `dvce_created_tstamp` | [Tracker](/docs/sources/trackers/index.md)                                          | The device's timestamp when the event was created                                                                                                                      | ✅             |
```
istreeter, line 14: I know that below you describe the inaccuracies in this timestamp.  But to end users I think this is a really important point to understand.  Maybe it should be up here in the table too, in case users don't read the detail below the table.

Same for `dvce_sent_tstamp`.

docs/events/timestamps/index.md
```diff
@@ -0,0 +1,33 @@
...
+| `refr_device_tstamp`  | [Enrich](/docs/api-reference/enrichment-components/index.md)                        | Timestamp extracted from the [cross-domain navigation query string](/docs/pipeline/enrichments/available-enrichments/cross-navigation-enrichment/index.md), if present | ❌             |
+| `load_tstamp`         | [Loader](/docs/destinations/warehouses-lakes/loading-process/index.md) or warehouse | Timestamp when the event was loaded into the warehouse                                                                                                                 | ✅             |
+
+The `load_tstamp` is added either by the Loader or by the warehouse at the point of loading, depending on the Loader/warehouse. Use this timestamp for incremental processing.
+
+## Derived timestamp
+
+Snowplow recommends using the `derived_tstamp` as the primary event timestamp for analysis. It allows for devices with incorrectly set clocks, or delays in event sending due to network outages.
```
istreeter, line 27: This might be true (not my area!) but a conflicting point is that downstream analysis should pay attention to how the warehouse table is partitioned.

For example, if the warehouse is partitioned by `load_tstamp`, then it is really important to use clauses like `WHERE load_tstamp > ?`.  I don't think we should give the impression that `derived_tstamp` is the one and only timetsamp to use in analysis.

istreeter (approved):
Very nice addition to the docs.  Thanks!

docs/events/timestamps/index.md
```diff
@@ -0,0 +1,39 @@
...
+
+The `derived_tstamp` corrects for inaccurate timestamps and delays.
+
+It's calculated as `collector_tstamp` minus the difference between the `dvce_sent_tstamp` and the `dvce_created_tstamp`. Alternatively, if `true_tstamp` is available, `derived_tstamp` is the same as `true_tstamp`.
+
+The calculation has two assumptions:
+* We assume that, although `dvce_created_tstamp` and `dvce_sent_tstamp` can both be inaccurate, they're inaccurate in precisely the same way. If the device clock is 15 minutes fast at event creation, then it remains 15 minutes fast at event sending, whenever that might be.
+* We assume that the time taken for an event to get from the device to the Collector is neglible, i.e. the lag between `dvce_sent_tstamp` and `collector_tstamp` is 0 seconds.
+
+## What timestamp to use?
```
agnessnowplow, line 35: I would call it out that it depends on how you plan on using the data. If you want to use it to filter on the `atomic.events` table directly then for sure they would need to use the timestamp it is partitioned on. For actual reporting, the `derived_tstamp` is advised to be used throughout (this is what we default as `start_tstamp` in our packaged modelled derived tables). The rest are for custom use cases or for debugging (?) to understand more on how the data is formed/sent/integrated into our pipeline/their warehouse etc.

agnessnowplow (approved):

PR #1263, opened 2025-05-29
Sync with Unity tracker API changes

jethron:
Document some missing parameters/defaults for the Unity tracking SDK as discovered by @snow-leosanches.

Let us know if you find anything else missing Leo!


matus-tomlein (approved):
Thanks Jethro!

PR #1290, opened 2025-06-19
Adding aggregated engagement tracking to snowplow.js

jrpeck1989:


snowplow.js
```diff
@@ -43,6 +78,7 @@ const createTrackerConfig = (cookieName) => {
     cookieDomain: `.${domain[1]}.${domain[0]}`,
     cookieName,
     cookieSameSite: 'Lax',
+    keepalive: true,
```
mscwilson, line 81: what's this do?

snowplow.js
```diff
@@ -64,16 +100,18 @@ const createTrackerConfig = (cookieName) => {
 const setupBrowserTracker = () => {
   newTracker(
     'snplow5',
-    'https://collector.snowplow.io',
+    // 'https://collector.snowplow.io',
```
mscwilson, line 103: fix these commented out bits please

mscwilson (approved):
needs tidying up from testing but otherwise lgtm

jrpeck1989:
> what's happening with this PR?

I need to have a meeting with Ostap to go over the implementation and verify it won't affect other tracking that is already in place

alexanderdean:
Nudge @pif

pif (commented):
Is the goal to enable aggregated tracking event on `biz1` pipeline or just on the `sales` pipeline?

Also please uncomment `biz1`.

snowplow.js
```diff
@@ -84,18 +122,49 @@ const setupBrowserTracker = () => {
   addGlobalContexts([selectedTabContext])
 
   enableLinkClickTracking()
+
   enableActivityTracking({
     heartbeatDelay: 10,
     minimumVisitLength: 10,
   }) // precise tracking for the unified log
 
+  enableActivityTrackingCallback({
```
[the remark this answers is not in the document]
jrpeck1989: > Is the goal to enable aggregated tracking event on biz1 pipeline or just on the sales pipeline?
>
> Also please uncomment biz1.

Just the sales pipeline - the goal is an isolated trial to size the dataset and judge whether the aggregated method is worth recommending to customers, without touching production tracking. `biz1` is restored (it was a local-testing artifact, my apologies). I've also taken a different approach to keep the experiment fully isolated: rather than re-pointing `snplow5`, I've added a dedicated sales tracker pointing at `com-snplow-sales-aws-prod1.collector.snplow.net`. `snplow5` and `biz1` are back to their production collectors and behave exactly as before.

> Does this mean it would send aggregated pings instead of the regular page_ping? Or that would be in addition to the regular page_pings?

In addition - they don't interfere. I checked the tracker core: `enableActivityTracking` (the regular `page_ping`) and `enableActivityTrackingCallback` (the aggregation) register under separate config keys (`pagePing` vs `callback`), so both run independently on the same tracker.

In this PR though, I've scoped them to different pipelines: production keeps `page_ping` as-is, and the sales tracker gets only the aggregated event (plus page views). So nothing changes for the regular `page_ping` data on `snplow5`/`biz1`.

> I think we just use these 2 columns from the regular unified views dbt project:
> 
>    engaged_time_in_s,
>    absolute_time_in_s,
> Maybe we can do smth like this to scope it per-tracker? Similar to how we enable special things per-tracker on the main website:
> 
> ...window.snowplow("enableActivityTrackingCallback:eng.aws-dev1", ...);

Done - scoped per-tracker exactly as you suggested. Since we initialise via the `@snowplow/browser-tracker` module API (not the `sp()` tag API), the equivalent is the optional trackers array argument:

```
enableActivityTrackingCallback({ ... }, ['sales'])
trackSelfDescribingEvent({ ... }, ['sales'])
```

and the regular `page-ping` activity tracking + clicks/forms are scoped to ['snplow5', 'biz1']. Net result: sales receives only `page_view` + `page_view_aggregated`.

On the columns - good point. The aggregated event captures `activeSeconds` (engagement proxy) and the scroll-offset min/max. `activeSeconds` is the analogue of `engaged_time_in_s`; I haven't included an `absolute_time_in_s` equivalent yet. I think this is fine though as the existing time tracking will continue to flow to the prod pipeline as normal. I'll work on a different data model on the new columns separate from your main modelling jobs.

PR #1283, opened 2025-06-13
Snowplow cli mcp intro

jamessnowplow:
It's kind of dry, happy to make it a bit more colourful if anyone has any ideas.

PR #1294, opened 2025-06-23
Delete recipes section

mscwilson:
Most of these recipes were over 3 years old. As of #1283 the two recent ones have been migrated to Tutorials, so this section can be removed.


stanch (approved):
Amen.

PR #1295, opened 2025-06-23
Tidy up tutorials

mscwilson:
* Added an intro
* Cleaned up the topics list
* Standardised the headings and blurbs


johnmicahreid:
I don't like the way the description appears, it takes up way too much space at the start of the page and pushes content below the fold. Either we should make it a single line (and not break out the categories - they are quite self-explanatory), or condense it into a tooltip: 
Screenshot 2025-06-23 at 14 56 22

PR #1197, opened 2025-04-07
Initial Signals Docs

Jack-Keene:

PR #1302, opened 2025-06-26
Update Signals pages

mscwilson:
Restructured and updated the main Signals pages


docs/signals/concepts/index.md
```diff
@@ -4,80 +4,224 @@ sidebar_position: 1
...
-The foundation of Signals are entities (distinct from [Snowplow entities](/docs/fundamentals/entities/index.md), but they can be related).
+```mermaid
+flowchart TD
+    Entity --> Stream
+    Entity --> Batch
+    Entity --> Int1[**Intervention 1**]
+    Entity --> Int2[**Intervention 2**]
 
-An <dfn>entity</dfn> is the core unit that all the other Signals features relate to, and can be anything with an "identifier" that you can capture in a Snowplow event.
+    subgraph Stream
```
matus-tomlein, line 18: I am not sure if this is technically correct – individual attributes have no relation to entities and also don't specify whether they should be processed in batch or stream. This is defined at the View level. I think it would be more correct to link the entities to views and also have the batch/stream distinction on the views, not on the attributes

docs/signals/concepts/index.md
```diff
@@ -4,80 +4,224 @@ sidebar_position: 1
...
+
+Choose the source which fits your use case. For example, the attribute `last_product_viewed` is best calculated from events in-stream during a session, while `num_views_in_last_7_days` is best calculated from historical events.
+
+An example configuration for a view based on a user entity:
+
+```mermaid
+flowchart TD
+    User[User: `user_id`] --> Stream
+
+    subgraph Stream[Stream attributes]
```
matus-tomlein, line 139: Also here it's a bit confusing to me that attributes are linked to the User entity and are defined as "Stream" – this is all part of the View definition, attributes don't define that, they are reusable across entities or stream/batch.

docs/signals/concepts/index.md
```diff
@@ -4,80 +4,224 @@ sidebar_position: 1
...
+This view could be imagined like this as a table once the attributes are calculated:
+
+| `user_id` | `number_of_pageviews` | `last_product_viewed` |
+| --------- | --------------------- | --------------------- |
+| `abc123`  | 5                     | `"Red Shoes"`         |
+| `def456`  | 10                    | `"Blue Hat"`          |
+
+You can use views individually in your application to retrieve attributes, or combine them into services.
+
+### Services group views
```
matus-tomlein, line 156: This is a short term limitation that services can only group views, they should be able to pick and choose individual attributes from different views (e.g., `num_page_views` from `view1` and `total_value` from `view2`).

I think the page `signals/configuration/views-services/` gives a better distinction between views and services – defining vs consuming attributes. Maybe we could somehow combine the explanation here and there?

docs/signals/configuration/views-services/index.md
```diff
@@ -1,21 +1,65 @@
...
+* Iterate on attribute definitions without worrying about breaking downstream processes
+* Migrate to new view versions by updating the service definition, without having to update the application code
 
+## Attribute management
 
-### Defining a view
-You can define a `View` by passing in a list of previously defined attributes. Here's an example:
+TODO this is very confusing
+
+Every view has a version number. If you change the attributes of a view, increase the version number.
```
matus-tomlein, line 23: Just noting that we need to expand on this – you can also change attributes of a view without changing the version number, the question is when to change the version number (what is a breaking change). We have a task in the backlog on this.

docs/signals/configuration/views-services/index.md
```diff
@@ -1,21 +1,65 @@
...
-A `Service` is a collection of views that streamlines the retrieval of multiple views. By grouping related views into a `Service`, you can efficiently manage and access user insights, making it easier to personalize applications and analyze behavior.
+By using services you can:
+* Iterate on attribute definitions without worrying about breaking downstream processes
+* Migrate to new view versions by updating the service definition, without having to update the application code
 
+## Attribute management
 
-### Defining a view
-You can define a `View` by passing in a list of previously defined attributes. Here's an example:
+TODO this is very confusing
```
matus-tomlein, line 21: Agree that this is hard to understand, I think we should separate the different cases and not try to explain "views with attributes" (those that are calculated in the stream or batch engine) and "views based on existing customer tables" (those where the user defines their own table and then provides a data source for it). The data sources are only necessary to define in the latter case, in the former (which is what Signals is really about), data sources should be handled automatically.

docs/signals/configuration/stream-calculations/index.md
```diff
@@ -0,0 +1,42 @@
...
+        my_service,
+        my_other_service
+    ])
+```
+
+Once applied, Signals will calculate the attributes within these views or services from the Snowplow event stream.
+
+If you only want to publish the attribute definitions, and don't want to calculate the attribute values now, use the `online=False` [view configuration option](/docs/signals/configuration/views-services/index.md).
+
+## Stream volume limit
```
matus-tomlein, line 36: This is only partly true – the differences are more nuanced. I have added a commit in the other PR with an overview table comparing the limitations of stream vs batch processing (sorry I made a conflict for your PR): https://github.com/snowplow/documentation/pull/1197/commits/05cabb6e50ee8a3ef826cf8c7a22feb32e732c39

I am not sure if this is the best place to put the discussion of limitations of stream vs batch, I think I would prefer if this and the next page were grouped under a common top level page giving an introduction to the two types of processing and discussing their limitations as well. I think we are missing some overview page here before jumping into the options.

PR #1314, opened 2025-07-09
Snowbridge 3.3.0

colmsnowplow:
This PR:

- Adds documentation of new features for Snowbridge 3.3.0 release (in rollout now), and updates the componentVersion.
- Updates all 'reference' type codeblocks to use a reference to the version specified, instead of master (our code examples were out of sync with the documented version before this)
- Fixes an issue where some files then had duplicated headers for some reason
- Removes a duplicate reference to `hcl` in the languages array.

That last change is an attempt to get codeblock formatting to work for hcl, but it appears still not to work. @mscwilson would you have any ideas on that?

It works for js reference code blocks, and it doesn't work for hcl reference or non-reference blocks. I've tried to find why but I'm lost.

(note: not all the features in 3.3.0 are documented here, because those are mostly for our internal under-the-hood use in event streaming, not really relevant for typical customers or community users)

### DIFF
--- a/docs/destinations/forwarding-events/snowbridge/configuration/index.md
+++ b/docs/destinations/forwarding-events/snowbridge/configuration/index.md
@@ -4,6 +4,11 @@ date: "2022-10-20"
 sidebar_position: 300
 ---
 
+```mdx-code-block
+import {versions} from '@site/src/componentVersions';
+import CodeBlock from '@theme/CodeBlock';
+```
+
 # Configuration Overview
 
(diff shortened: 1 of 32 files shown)


colmsnowplow:
This might have something to do with the hcl formatting issue I guess:

```
yarn list prism-react-renderer
yarn list v1.22.22
warning Filtering by arguments is deprecated. Please use the pattern option instead.
├─ @docusaurus/theme-classic@2.4.3
│  └─ prism-react-renderer@1.3.5
├─ @docusaurus/theme-common@2.4.3
│  └─ prism-react-renderer@1.3.5
└─ prism-react-renderer@2.4.1
✨  Done in 0.23s.
```

stanch (approved):

PR #1316, opened 2025-07-10
Fix italics fonts

phil-snowplow:
Italic text wasn't rendering with true italics (i.e. built into the font file), but instead was using synthetic italics generated by the browser. Updated the `@font-face` declarations in `src/css/custom.css` to fix this


mscwilson (approved):

AH-Avalanche:
Yeah, we can close it, make sure the italics look good with the new variable font. Looks like they do

PR #1322, opened 2025-07-14
Fix JSON error when fetching API key locally

mscwilson:
There's a function in `Root.js` to fetch the Product Fruits API key. The response is expected to be a short string.

Weirdly, when running locally it fetches some HTML instead. This PR adds a check that the response is valid.


mscwilson:
this now catches the error so it's logged in the console, it won't bring up a big annoying box over the screen anymore

PR #1333, opened 2025-07-22
Update schema properties component

phil-snowplow:
This PR redesigns the `SchemaProperties` component with a cleaner, card-based one that's based on the Avalanche design system's [panel](https://github.com/snowplow-product/msc-frontend-service/blob/develop/src/components/organisms/Panel.tsx) component. The new layout makes the schema information easier to read and fixes some overflow issues the old component had.

**Key Changes:**

- Improved layout: Replaced the old admonition layout with a card for better visual hierarchy.
- Renamed sections: "Example", "Schema", and "Warehouse query" are renamed to be simpler and more concise (and remove emojis).
- Simplified schema table layout: Made the **Schema** table view a two-column layout instead of four columns to better fit the narrow space.
- Consistent Styling: Uses Tailwind/Shadcn for consistent styling.

**Note:** I'm currently in-lining the `panel` component into `SchemaProperties`. We can break that out into it's own panel component if we see a need to make it reusable.
 
**Old:**
CleanShot 2025-07-22 at 16 46 40

**New:**
CleanShot 2025-07-22 at 16 45 18

### DIFF
--- a/src/components/SchemaProperties/index.js
+++ b/src/components/SchemaProperties/index.js
@@ -37,11 +37,11 @@ export default function SchemaProperties(props) {
           </div>
           {props.overview && (
             <div className="gap-2 text-base leading-none font-bold flex flex-col items-start">
-              <span>Tracker Compatibility</span>
+              <span>Tracker compatibility</span>
               <div className="flex gap-4">
                 <span className="text-base font-normal text-muted-foreground flex-initial">Web: {props.overview.web ? '✅' : '❌'}</span>
                 <span className="text-base font-normal text-muted-foreground flex-initial">Mobile: {props.overview.mobile ? '✅' : '❌'}</span>
-                <span className="text-base font-normal text-muted-foreground flex-initial">Tracked Automatically: {(typeof props.overview.automatic === 'string') ? props.overview.automatic : (props.overview.automatic ? '✅' : '❌')}</span>
+                <span className="text-base font-normal text-muted-foreground flex-initial">Tracked automatically: {(typeof props.overview.automatic === 'string') ? props.overview.automatic : (props.overview.automatic ? '✅' : '❌')}</span>
               </div>
             </div>
           )}


src/components/SchemaProperties/index.js
```diff
@@ -0,0 +1,124 @@
...
+      </div>
+      <div className="p-6 pt-0 w-full">
+        <div className="space-y-4">
+          <div className="gap-2 text-base leading-none font-bold flex flex-col items-start">
+            <span>Schema URI</span>
+            <span className="text-muted-foreground leading-snug font-normal">iglu:{props.schema.self.vendor}/{props.schema.self.name}/{props.schema.self.format}/{props.schema.self.version}</span>
+          </div>
+          {props.overview && (
+            <div className="gap-2 text-base leading-none font-bold flex flex-col items-start">
+              <span>Tracker Compatibility</span>
```
mscwilson, line 40:
```suggestion
              <span>Tracker compatibility</span>
```

src/components/SchemaProperties/index.js
```diff
@@ -0,0 +1,124 @@
...
+              <div className="flex gap-4">
+                <span className="text-base font-normal text-muted-foreground flex-initial">Web: {props.overview.web ? '✅' : '❌'}</span>
+                <span className="text-base font-normal text-muted-foreground flex-initial">Mobile: {props.overview.mobile ? '✅' : '❌'}</span>
+                <span className="text-base font-normal text-muted-foreground flex-initial">Tracked Automatically: {(typeof props.overview.automatic === 'string') ? props.overview.automatic : (props.overview.automatic ? '✅' : '❌')}</span>
```
mscwilson, line 44:
```suggestion
                <span className="text-base font-normal text-muted-foreground flex-initial">Tracked automatically: {(typeof props.overview.automatic === 'string') ? props.overview.automatic : (props.overview.automatic ? '✅' : '❌')}</span>
```

PR #1337, opened 2025-07-23
Add live Signals widget

mscwilson:
Adds a live display of some calculated session attributes for the current user.

Will need updating when the backend is deployed.

Screenshot 2025-07-23 at 11 51 44


mscwilson:
Closed because low priority

PR #1346, opened 2025-07-30
Right side table of content upgrade

AH-Avalanche:
- Reduced amount of arrows
- Arrows appear on hover or if menu item is active
- Better Active state
- Folder lines easier to follow for the user when they are in deeper sections


mscwilson (commented):
I agree, looking good but text could be bigger

mscwilson (approved):

phil-snowplow:
One (very) small nit: the text moves slightly to the right on hover, probably because the `border-left-width` changes from 0px to 1px

https://github.com/user-attachments/assets/0b78fbf9-777e-4795-98c5-5d1a8931aace

PR #1348, opened 2025-07-30
Add migration guides

johnmicahreid:
Adds migration guides for Segment and GA

### DIFF
--- a/docs/resources/migration-guides/google-analytics/index.md
+++ b/docs/resources/migration-guides/google-analytics/index.md
@@ -14,12 +14,12 @@ There are [significant differences](https://snowplow.io/comparisons/snowplow-vs-
 
 This table shows some key differences:
 
-| Feature                 | Google Analytics 4                                                                                | Snowplow                                                                                                                           |
-| ----------------------- | ------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------- |
-| Capturing user behavior | Uses `gtag('event', ...)` with event names, or `dataLayer.push()` with GTM triggers and variables | Direct tracker SDK calls for built-in or custom [event](/docs/fundamentals/events/index.md) types                                  |
-| Contextual event data   | Included in the event `parameters` object                                                         | Included in the event as reusable [entities](/docs/fundamentals/entities/index.md)                                                 |
-| Warehouse tables        | BigQuery only; daily partitioned `events_YYYYMMDD` table with nested fields                       | In BigQuery and Snowflake, one big table with columns for every event or entity; in Redshift, one table per custom event or entity |
-| Data validation         | Optional and limited schema enforcement                                                           | All events and entities defined by JSON schemas                                                                                    |
+| Feature                 | Google Analytics 4                                                                                | Snowplow                                                                                                                                                                   |
+| ----------------------- | ------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Capturing user behavior | Uses `gtag('event', ...)` with event names, or `dataLayer.push()` with GTM triggers and variables | Direct tracker SDK calls for built-in or custom [event](/docs/fundamentals/events/index.md) types                                                                          |
+| Contextual event data   | Included in the event `parameters` object                                                         | Included in the event as reusable [entities](/docs/fundamentals/entities/index.md)                                                                                         |
+| Warehouse tables        | BigQuery only; daily partitioned `events_YYYYMMDD` table with nested fields                       | In [most warehouses](/docs/destinations/warehouses-lakes/index.md) one big table with columns for every event or entity; in Redshift, one table per custom event or entity |
+| Data validation         | None                                                                                              | All events and entities defined by JSON schemas                                                                                                                            |
 
 ### Event structure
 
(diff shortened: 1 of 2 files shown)


docs/resources/migration-guides/segment/index.md
```diff
@@ -8,22 +8,7 @@ This guide helps technical implementers migrate from Segment to Snowplow.
 
 ## Platform differences
 
-There are a number of differences between Segment and Snowplow as a data platform.
```
johnmicahreid: I think this feature comparison table is actually quite important - we can probably make the text a bit less marketing-speaky and more neutral, but it's important to lay out the differences at a higher level.

docs/resources/migration-guides/segment/index.md
```diff
@@ -0,0 +1,157 @@
...
+Snowplow uses a single [`atomic.events`](/docs/fundamentals/canonical-event/index.md) table in warehouses like Snowflake and BigQuery. Events and entities are stored as structured columns within that table, simplifying analysis.
+
+## Migration phases
+
+We recommend using a parallel-run migration approach. This process can be divided into three phases:
+1. Assess and plan
+2. Implement and validate
+3. Cutover and finalize
+
+### Assess and plan
```
johnmicahreid, line 100: The section from here on out feels a bit too bullet-point-y - definitely AI-generated, but with not quite enough context or detail to actually be useful. We should assume that people reading this migration guide are not actually familiar with Snowplow, so taking a bit of extra time to explain concepts in more detail could be helpful. 

I also find myself wanting numbers? e.g. "1. Assess and plan", "1.1 Audit existing implementation" etc

docs/resources/migration-guides/segment/index.md
```diff
@@ -0,0 +1,157 @@
...
+| `group()`        | `group('company-123', {name: 'Acme Corp', plan: 'Enterprise'})` | No direct equivalent. Attach a custom `group` entity to your events, or track group membership changes as custom events with `group_joined` or `group_updated` schemas.      |
+| `alias()`        | `alias('new-user-id', 'anonymous-id')`                          | No direct equivalent. Track identity changes as custom events with `user_alias_created` schema. Use `setUserId` to update the current user identifier for subsequent events. |
+
+### Warehouse data structure
+
+Segment loads each custom event type into separate tables, for example, `order_completed`, or `product_viewed` tables. Analysts must `UNION` tables together to reconstruct user journeys.
+
+Snowplow uses a single [`atomic.events`](/docs/fundamentals/canonical-event/index.md) table in warehouses like Snowflake and BigQuery. Events and entities are stored as structured columns within that table, simplifying analysis.
```
johnmicahreid, line 91: I would also mention that we provide OOTB data models to model common analytics use-cases, and link to our DBT packages. Otherwise we risk this migration guide focusing solely on the upstream tracking bit, vs comparing one data platform to another.

docs/resources/migration-guides/segment/index.md
```diff
@@ -0,0 +1,157 @@
...
+- Monitor data quality and pipeline health
+- Update [dbt models](https://docs.snowplow.io/docs/modeling-data/modeling-your-data/dbt/) to use the new data structure
+
+### Cutover and finalize
+
+#### Update downstream consumers
+- Migrate BI dashboards to query Snowplow tables
+- Test all data-dependent workflows
+
+#### Configure integrations
```
johnmicahreid, line 150: I think we could say more here - both of these are full products in their own right, and ProfServ has mentioned that these have been quite tricky to do in the past.

docs/resources/migration-guides/segment/index.md
```diff
@@ -0,0 +1,157 @@
...
+- Export your complete Segment tracking plan, using one of these methods:
+  - Ideally, use the Segment Public API to obtain the full JSON structure for each event
+  - Manually download CSVs from the Segment UI
+  - Infer it from warehouse data
+
+#### Design Snowplow tracking plan
+- Translate Segment events into Snowplow self-describing events
+  - The [Snowplow CLI](/docs/data-product-studio/snowplow-cli/index.md) MCP server can help with this
+- Identify reusable entities that can replace repeated properties
+- Create JSON schemas for all events and entities
```
johnmicahreid, line 114: We should mention Data Products here - it is about more than just events and entities. And this isn't quite the right order of operations - the MCP server does actually create the JSON schemas.

johnmicahreid (commented):
@mscwilson overall this is better, and I like having the text shorter. It does feel a bit too terse and low-level at the moment though - it only focuses on tracking, and assumes a lot of knowledge about Snowplow. I wouldn't mind adding a bit more context about our platform, bearing in mind the likely audience is someone who is looking to understand the overall differences and what they should be thinking about.

docs/resources/migration-guides/segment/index.md
mscwilson: that depends what this document is for. I think overall comparisons of the platforms is a marketing thing, they can already read about that on our marketing website https://snowplow.io/comparisons/snowplow-vs-segment

we could link to that comparison?

this is about how to technically migrate - what concepts map to what, what do they need to action to make it work

johnmicahreid: yup, let's at least link to the comparison

docs/resources/migration-guides/google-analytics/index.md
```diff
@@ -0,0 +1,263 @@
...
+  event: 'purchase',
+  ecommerce: {
+    transaction_id: '12345',
+    value: 25.42,
+    currency: 'USD'
+  }
+});
+```
+
+In GTM, you would configure a Snowplow tag that fires when `event` equals `purchase`, mapping the ecommerce properties to a custom Snowplow self-describing event or using the built-in ecommerce tracking.
```
jrpeck1989: Or Snowplow's Ecommerce plugin

johnmicahreid:
LGTM, let's get these launched!

PR #1369, opened 2025-08-18
Consistent-ify bigquery/lake/snowflake loaders config references for common-streams

istreeter:
Our apps have recently bumped common-streams versions a few times, and we haven't always kept the config reference up to date.

Ideally, we would arrange this documentation to import a shared config reference snippet for all common-streams apps.  But that's out of scope for this PR.  I'm just doing the easy thing and triplicating the same config reference into all three loaders.


oguzhanunlu (approved):
thanks Ian!

PR #1389, opened 2025-09-04
Bug Fix Particle Banner DOM enhancement

AH-Avalanche:
### Why

-   The banner’s DOM-based particle simulation was CPU-intensive due to many per-frame DOM mutations running continuously even when not interacting or visible.

### What changed

-   Switched to canvas renderer: Added  src/components/ui/CanvasParticles.tsx to draw all particles in one  <canvas>, removing per-particle DOM nodes and style recalculations.

-   Paused when idle/offscreen: In src/components/ui/InteractiveBanner.tsx, particles only render when the banner is hovered, in viewport (IntersectionObserver), and tab is visible  (Page Visibility API).

-   Throttled frame rate: Simulation capped at  ~30fps.

-   Reduced work per frame: Fewer particles, softer physics, and trails disabled when idle.

-   Brand-aligned color: Particles now render using theme  --primary via  hsl(var(--primary)).

-   Cleanup: Removed  src/components/ui/PhysicsParticles.tsx and its import.

### Files

-   Updated: src/components/ui/InteractiveBanner.tsx

-   Added: src/components/ui/CanvasParticles.tsx

-   Removed: src/components/ui/PhysicsParticles.tsx

### Impact

-   Significant CPU reduction on idle pages and when the banner is offscreen.

-   Fewer layout/paint invalidations; smoother interactions.

-   Visuals preserved with primary brand color and subtle glow.


mscwilson (approved):
lgtm

miike:
It's down to about 15% now (it is still re-rendering the DOM without interaction to compute a gradient) ideally we'd want it somewhere below 1% (you can check in Chrome what the performance looks like compared to your other tabs).

PR #1391, opened 2025-09-09
Upgrade to Docusaurus v3

mscwilson:
Upgrade to Docusaurus v3.9

Changes:
- Updated and added various dependencies
- Updated the Tutorials UI
- Changed `:::caution` admons to `:::warning` because Caution isn't supported
- Removed the Runnable codeblocks and the custom swizzled `CodeBlock/Content` component
- Updated various other Docusaurus components 
- Fixed some tables that weren't formatted correctly
- Escaped `{` in code blocks
- Removed `mdx-code-block` blocks for some imports - imports can be written directly in the file
- Tables and some paragraphs that use `{version.whatever}` now do so directly rather than as a `ReactMarkdown` component

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,10 +17,10 @@
   "dependencies": {
     "@algolia/client-search": "^5.0.0",
     "@braintree/sanitize-url": "^6.0.1",
-    "@docusaurus/core": "^3.9.0",
-    "@docusaurus/plugin-client-redirects": "^3.9.0",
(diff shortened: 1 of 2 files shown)


AH-Avalanche:
@mscwilson, I have updated the code block component to resolve the issue. Buttons on code blocks are overlaying the code instead of being above it. We had to integrate a whole new component to gain better control, and now we have full control over it. Let me know what you think.

PR #1394, opened 2025-09-09
Fix UI bug in tab groups

mscwilson:
The relevant line is 820/832. VS Code autoformatted the file. I can change it back if this is a problem

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ manifests
 
 # Local Netlify folder
 .netlify
+.claude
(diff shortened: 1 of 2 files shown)


phil-snowplow:
Good find! While we're at it, I fixed a related issue where the contents of a tab shift slightly on hover. This was because it was set to a default 2px transparent border, but then changed to a 1px border on hover ([code](https://github.com/snowplow/documentation/pull/1394/commits/7b1a7c5ce9f40dee4c618500a2c3dacdbc8ae62d#diff-44813f307e729042af7e70beff6f32ea63a7941bc100043a49e84d229ea2570fL839-R839)).

phil-snowplow (approved):

PR #1414, opened 2025-09-18
[DPS-1424] Update documentation for alerts to include absolute thresholds

DianaIgnatescu:


gibbok-snowplow (commented):
Thanks for the PR, few suggestions:
- rename the image with full name, so instead of dq please consider data_quality, this is better for SEO
- I would add few uses cases describing the value

docs/data-product-studio/data-quality/failed-events/monitoring-failed-events/alerts/failed-event-alerts/creating-alerts/index.md
```diff
@@ -92,6 +92,16 @@ When using "When above value" trigger:
 
 ![Value threshold configuration with insights panel](images/data_quality_absolute_threshold.png)
 
+#### Insights from historical data
```
gibbok-snowplow: A sligly modified version here, up to you if you would like to use it:

```

#### Insights from recent data



When you configure a threshold alert, the Insights panel displays data from the last 7 days to help you choose an appropriate threshold:



- **Average event volume** – Shows the average counts of valid and failed events (per relevant time unit).  

- **Suggested threshold** – Provides a recommended value based on recent patterns.  

- **Apply suggestion** – Click to insert the suggested value into the threshold field.  



Use these insights to set realistic thresholds that reflect your actual event patterns.



**Alert delivery:** If the trigger condition is met, the alert is sent at most once per delivery interval you selected.

```

gibbok-snowplow (approved):

PR #1415, opened 2025-09-19
Add instructions on how to authenticate using Signals Sandbox through the SDK

matus-tomlein:
Adds snippets to explain how to authenticate using the Signals Sandbox token through the SDK.

Not sure if we need more context for explaining Signals Sandbox or that will be explained separately? cc @johnmicahreid


igneel64:
In my opinion this should only belong in the sandbox site, not the main docs.
Or in a completely separate section of the sandbox.

PR #1423, opened 2025-09-26
Web Tracker - Screen Tracking Plugin - Update to use correct code references

jkersu:
Commands in doc do not match function names in the plugin code


tobiasrask (approved):
Looks good, thanks! 💎

PR #1435, opened 2025-10-02
Add Identities docs

mscwilson:
Documentation for the new Identities product.


docs/identities/index.md
```diff
@@ -0,0 +1,64 @@
...
+
+Your Identities infrastructure is deployed into the same cloud as your pipeline. The core components are:
+* **Identities API**: used for identity operations
+* **Managed Postgres database**: persists the identity graph and performs graph operations. Snowplow uses [Aurora](https://aws.amazon.com/rds/aurora/) on AWS and [AlloyDB](https://cloud.google.com/products/alloydb) on GCP.
+
+## Using Identities
+
+Deploy and configure Identities using [Snowplow Console](https://console.snowplowanalytics.com).
+
+Steps for using Identities:
```
jborlase-snowplow: Do we need these steps, only to launch directly into the steps?

docs/identities/index.md
```diff
@@ -0,0 +1,64 @@
...
+### 2. Configure Identities
+
+Use [Console](https://console.snowplowanalytics.com) to configure Identities with the identifiers you've chosen.
+<!-- TODO: do we have any info about priorities? -->
+
+Set optional filters to specify which events will be sent to Identities for resolution. Events that are filtered out will not have an [identity entity](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/identity/jsonschema/1-0-0) and corresponding `snowplowId` attached to them.
+
+### 3. Use your new identity data
+
+Set up the Identities dbt package to create your new identity tables. The existing Snowplow dbt models support Identities. Use your identity data to gain a better understanding of your users.
```
jborlase-snowplow: Existing Snowplow dbt models do not support? We shouldn't call that out - just the fact we have the identities dbt package. 

Lets also want to call out how the identifiers can be used for real-time usecases as well as in the warehouse.

docs/identities/concepts/index.md
```diff
@@ -0,0 +1,386 @@
...
+```
+
+Identities is based on several core concepts.
+* **Identifiers** are the properties in the event payload that correspond to a user
+* **Profiles** are collections of linked identifiers that represent a single user
+* **Merges** occur when identifiers link two previously separate profiles together
+
+## Identifiers
+
+Identifiers are the key/value pairs that Identities extracts from events and uses to resolve identity. Each identifier has a name or type, such as `domain_userid`, and a value that's the actual ID in the event payload, such as `a43eb2f1-...`.
```
jborlase-snowplow: "Or type"?

phil-snowplow: In the UI mocks, we currently use `Identifier Name`, so you you could probably remove the "or type"

PR #1706, opened 2026-04-07
Keep referrer for external links

mscwilson:
## What changed?

Adds a client module to remove `noreferrer` for external links.

Why a client module rather than another solution:
`@docusaurus/Link` isn't swizzleable.
The closest is to swizzle `MDXComponents/A` for external links within markdown pages, and then separately handle external links in the footer (swizzle `Footer/LinkItem`) and custom React components.

## Why?

To help with data analysis, and to let external sites know we're driving traffic for them.


src/js/snowplowReferrer.js
```diff
@@ -0,0 +1,22 @@
+/**
+ * Client module that strips `noreferrer` from all external links.
+ * Keeps `noopener` for security. Uses Docusaurus's onRouteDidUpdate
+ * lifecycle to run once per page navigation.
```
stanch, line 4: So it does this at runtime? Is there a way to do this at build time?

PR #1723, opened 2026-04-13
Add noindex flag to link pages again

mscwilson:
Link pages were showing up in Algolia search results again. They just looked like empty pages - confusing.

I added `noindex` in #1503, but there was a regression in #1629. This PR fixes the behaviour, plus a comment explaining what's happening.

Also adds content to the link pages in case anyone ends up there.

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -155,12 +155,15 @@ const swapDocItemsToLinkItems = (generatedDocs, originalDocs) => {
       descriptions[docItem.id] = docItem.frontMatter.description
     }
     if (docItem.frontMatter.type === 'link') {
+      // Mutate the original frontMatter so useDoc() picks up noindex at render time.
+      // This is intentional — the spread into linkItems below copies the updated reference.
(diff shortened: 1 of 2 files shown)


sidebars.js
```diff
@@ -155,12 +155,15 @@ const swapDocItemsToLinkItems = (generatedDocs, originalDocs) => {
       descriptions[docItem.id] = docItem.frontMatter.description
     }
     if (docItem.frontMatter.type === 'link') {
+      // Mutate the original frontMatter so useDoc() picks up noindex at render time.
+      // This is intentional — the spread into linkItems below copies the updated reference.
+      docItem.frontMatter.sidebar_custom_props = {
```
stanch, line 160: `docItem.frontMatter.sidebar_custom_props.noIndex = true` ?

stanch: Ah, would not work if it’s undefined

stanch (approved):

PR #1745, opened 2026-04-21
[DPS-1823] Update autogen data models SS still referencing data products

DianaIgnatescu:
## What changed?

Updated screenshots in autogen data models.

## Why?

Some were still referencing data products instead of tracking plans.


gibbok-snowplow (approved):
LGTM thanks

PR #1749, opened 2026-04-22
Document tracker support for web atomic event properties

mscwilson:
## What changed?

Adds details about how to track certain atomic properties, because it wasn't clear. Now all the event properties that can be provided via trackers are explained in the OOTB events section.

Also clarifies that tracker-provided values have precedence over pipeline ones.


jborlase-snowplow (approved):
LGTM

PR #1752, opened 2026-04-27
Enrich 6.11.0

istreeter:


docs/api-reference/enrichment-components/configuration-reference/index.md
```diff
@@ -51,6 +51,10 @@ To accept the terms of license and run Enrich, set the `ACCEPT_LIMITED_USE_LICEN
...
 | `jsAllowedJavaClasses` (since *6.6.0*)                    | Optional. Default: `["*"]`. List of Java classes that the [JavaScript enrichment](/docs/pipeline/enrichments/available-enrichments/custom-javascript-enrichment/index.md) is allowed to access. This affects both `new <class>` and `<class>.<static method>` usage. Examples: `["java.lang.String", "java.net.URL"]`, `["java.net.*"]`. By default, all classes are allowed.                                                                                                              |
+| `compression.enabled` (since *6.11.0*)                    | Optional. Default: `false`. Enable compression on the enriched output stream. This is a preview feature; not all stream consumer applications support it yet. |
```
stanch: Should we list the consumers that support it?

istreeter: I do think it's a good idea to have a list, so that self-hosted customers can check the list.  But, currently that list will be empty!  We haven't released any of the decompressing applications yet.

Do you think the list belongs here, or in a separate page/section about compression?

stanch (approved):

PR #1769, opened 2026-05-07
Fix elasticsearch loader prometheus.tags description

oguzhanunlu:
The /metrics endpoint is unconditionally exposed by the health probe in common-streams; setting tags does not enable it. Tags only add common labels.


istreeter:
We could also use this opportunity to add elasticsearch loader to the list of consumers with support for compression?

oguzhanunlu:
I had to rebase and force push to get latest master.

istreeter (approved):

PR #1776, opened 2026-05-12
Show column name for events and entities

mscwilson:
Updates the `SchemaProperties` component to show the expected column name for that schema. Makes it easier for analysts to find where data has come from.

### DIFF
--- a/docs/reusable/event-query/_index.md
+++ b/docs/reusable/event-query/_index.md
@@ -11,8 +11,6 @@ export function fieldName(props, wh) {
         props.name +
         '_' +
         (
-          wh == 'bigquery' ?
-          props.version.replaceAll('-', '_') :
           props.version.split('-')[0]
         )
     );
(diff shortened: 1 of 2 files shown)


docs/reusable/event-query/_index.md
```diff
@@ -74,18 +93,15 @@ where
 
 </TabItem>
 
-<TabItem value="redshift" label="Redshift & Postgres">
+<TabItem value="synapse" label="Synapse Analytics">
```
stanch: Let’s remove Synapse altogether

src/components/SchemaProperties/index.js
```diff
@@ -14,6 +14,16 @@ export default function SchemaProperties(props) {
   const description = props.info || props.schema.description
   const hasProperties =
     props.schema.properties && Object.keys(props.schema.properties).length > 0
+  const warehouseNote = (
+    <p>
+      Redshift and older versions of the BigQuery Loader use different names —
```
stanch: Snowflake also uses different names, we turn schema fields into camelCase

PR #1780, opened 2026-05-13
Update Claude instructions

mscwilson:
* Clarifies a couple of instructions that it might have been confused by
* Adds an instruction to use the `SchemaProperties` component for schemas
* Makes the PR review workflow more explicit about what files to read
* Adds a validation check for external links


jborlase-snowplow (approved):
Nice change!

PR #1783, opened 2026-05-13
Replace cookie consent provider

mscwilson:
Replacing cookie-though with Ketch


src/components/ManageCookiePreferences/index.js
```diff
@@ -1,47 +1,55 @@
...
+    return {
+      analytics: parsed.analytics?.status === 'granted',
+      advertising: parsed.targeted_advertising?.status === 'granted',
+    }
+  } catch {
+    return null
+  }
+}
+
+const consentLabel = (consent) => {
```
gibbok-snowplow, line 19: please consider adding a test.
nit: you could use a .reduce

src/components/ManageCookiePreferences/index.js
```diff
@@ -1,47 +1,55 @@
...
-          'You currently have Necessary and Analytics cookies enabled'
-        )
-      } else {
-        setConsentState('You currently have only the Necessary cookies enabled')
-      }
-    }
+    setConsent(getConsentState())
+
+    const handler = () => setConsent(getConsentState())
+    window.ketch('on', 'consent', handler)
```
gibbok-snowplow, line 41: what if window.ketch is undefined?

src/pages/cookie-preferences.md
```diff
@@ -0,0 +1,51 @@
...
+| `_ketch_consent_v1_` | Stores your explicit cookie preferences after you interact with the consent banner.           |
+| `_swb`               | Set by Ketch to identify your browser session for consent management purposes.                |
+| `_swb_consent`       | Stores your current consent state, including the legal basis applied to each cookie category. |
+
+### Data analytics
+
+| Cookie                        | Description                                                                                                                                                                                               |
+| ----------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `_ga`                         | Installed by Google Analytics. Calculates visitor, session, and campaign data and tracks site usage. Stores information anonymously and assigns a randomly generated number to recognize unique visitors. |
```
gibbok-snowplow, line 40: just to double check... is Google Analytics installed?

snowplow.js
```diff
@@ -109,30 +110,40 @@ const setupBrowserTracker = () => {
...
-          Cookies.remove('sp')
-          reloadOnce()
-        }
+  let consentInitialized = false
+
+  window.ketch('on', 'consent', (consent) => {
+    const analyticsEnabled = consent.purposes?.analytics === true
+
+    if (!consentInitialized) {
+      consentInitialized = true
```
gibbok-snowplow, line 119: why do you need tit to set to true here? are you going to use consentInitialized in some other place?

src/pages/cookie-preferences.md
mscwilson: yes!

snowplow.js
```diff
@@ -109,30 +110,40 @@ const setupBrowserTracker = () => {
...
+      return
+    }
+
+    if (analyticsEnabled) {
+      disableAnonymousTracking({
+        stateStorageStrategy: 'cookieAndLocalStorage',
+      })
+      trackPageView()
+    } else {
+      const cookieKeys = document.cookie
```
gibbok-snowplow: should we use js-cookies as you did before even here?

snowplow.js
```diff
@@ -109,30 +110,40 @@ const setupBrowserTracker = () => {
...
+        .split(';')
+        .reduce((ac, str) => [...ac, str?.split('=')[0].trim()], [])
+      const snowplowCookies = cookieKeys.filter((cookieKey) =>
+        cookieKey.startsWith('_sp5_')
```
gibbok-snowplow: nit: you could have a const file for _sp5_ and place it there, so all these are in a single place

src/components/ManageCookiePreferences/index.js
mscwilson: it will always be defined https://github.com/snowplow/documentation/blob/0b1332dc2bb5b8253208b769621155f475fe5457/docusaurus.config.ts#L282

snowplow.js
mscwilson: users can change their cookie preferences without reloading the page

src/components/ManageCookiePreferences/index.js
mscwilson: we don't have tests in the docs 😔

gibbok-snowplow (approved):
LGTM

PR #1793, opened 2026-05-20
Vercel Signals Agents Accelerator updates

Jack-Keene:
## What changed?
- We should not be using an application wide parent component but more of a sibling like here [next.js/examples/with-segment-analytics/app/layout.tsx at canary · vercel/next.js](https://github.com/vercel/next.js/blob/canary/examples/with-segment-analytics/app/layout.tsx) 
- Remove service name as an env variable SNOWPLOW_SIGNALS_SERVICE_NAME as env variable.
- Create the getters in the JS tracker for pulling the data from a page 
 
## Why?

Accelerators should follow Snowplow/Next.js conventions


tutorials/signals-ai-agent-context/build-ai-integration.md
```diff
@@ -285,21 +284,20 @@ export function ChatWidget() {
...
-    <ChatWidget />
-  </SnowplowProvider>
+  <SnowplowTracker />
+  {children}
+  <ChatWidget />
 </body>
 ```
 
-The `ChatWidget` is placed inside `SnowplowProvider` to ensure the tracker is always initialized before the chat reads the session ID.
+The `ChatWidget` sits alongside `SnowplowTracker` in the layout. If the tracker hasn't initialized yet when the chat sends its first request, `getDomainSessionId()` returns an empty string and the agent responds without Signals context.
```
igneel64: > If the tracker hasn't in

Is this a case that can actually happen ? _(with an expected normal page load)_
If not let's remove the it, it takes away from the main point.

tutorials/signals-ai-agent-context/configure-signals.md
```diff
@@ -45,7 +45,7 @@ A [service](/docs/signals/concepts/#services) provides a pull-based API endpoint
 
 Services allow you to combine multiple attribute groups if needed, but for this tutorial, use just the one you created in the last step.
 
-Use this exact service name. It's the same as the `SNOWPLOW_SIGNALS_SERVICE_NAME` environment variable you configured in the Next.js application.
+You'll use this service name in your application code to fetch attributes from this service.
```
igneel64: i think this is not needed now.

igneel64 (approved):

PR #1820, opened 2026-06-09
Merge tracking CLI into the snowplow-cli

gleb-lobov:


gleb-lobov:
@claude I've addressed the capitalisation

docs/api-reference/snowplow-cli/data-structures/index.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: "Manage data structures via Snowplow CLI"
 sidebar_label: "Data structures"
-sidebar_position: 2
+sidebar_position: 20
```
cksnp, line 4: Is this a typo?

docs/api-reference/snowplow-cli/reference/index.md
```diff
@@ -2,7 +2,7 @@
 title: "Snowplow CLI command reference"
 sidebar_label: "Command reference"
 date: 2026-03-06
-sidebar_position: 1
+sidebar_position: 0
```
cksnp, line 5: Just out of curiosity - are they 0-indexed?

docs/api-reference/snowplow-cli/send-events/index.md
```diff
@@ -0,0 +1,88 @@
+---
+title: "Send events via Snowplow CLI"
+sidebar_label: "Send events"
+sidebar_position: 10
```
cksnp, line 4: There's clearly something I don't get about sidebar positions, so I won't comment any further in this PR :D

docs/api-reference/snowplow-cli/send-events/index.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Prerequisites
+
+You need [Snowplow CLI](/docs/event-studio/programmatic-management/snowplow-cli/index.md) installed. Sending events does not require Console credentials, so you can skip the configuration step if you only use the `events` subcommand.
+
+## Send an event
+
+The `events send` command sends a single [self-describing event](/docs/fundamentals/events/index.md#self-describing-events) to your Collector. You can either pass a complete self-describing JSON, or pass a schema URI and its data separately and let the CLI assemble the event for you.
+
+The only required flag is `--collector`, the domain of your Snowplow Collector. The examples below use the built-in `com.snowplowanalytics.snowplow/custom_event` schema, so they produce a valid event as long as you point `--collector` at a real Collector.
```
cksnp, line 22:
```suggestion
The only required flag is `--collector`, the hostname of your Snowplow Collector. The examples below use the built-in `com.snowplowanalytics.snowplow/custom_event` schema, so they produce a valid event as long as you point `--collector` at a real Collector.
```

cksnp (approved):
LGTM!

docs/api-reference/snowplow-cli/data-structures/index.md
gleb-lobov: Not really. Before the PR it was 1,2,3, and I've needed to add something between 1 and 2. Since I can't to 1,5, I have to move everything around to make space for the new page.

docs/api-reference/snowplow-cli/reference/index.md
gleb-lobov: I don't think there is any idexing, it's just integers ASC sorted. So you are free to put any values. Putting them with some space in between(10,20,30 instead of 1,2,3) allows the future pages to be injected in between, without changing the position of the existing page

docs/api-reference/snowplow-cli/send-events/index.md
gleb-lobov: I don't really know the difference, but quick googling seems to show that the domain might be correct here, WDYT?

https://runcloud.io/blog/hostname-vs-domain-name

cksnp: > I don't really know the difference, but quick googling seems to show that the domain might be correct here, WDYT?
> 
> https://runcloud.io/blog/hostname-vs-domain-name

You got me thinking. According to RFC 3986:
`authority = [ userinfo "@" ] host [ ":" port ]`
...so maybe the right term is just `host`?

cksnp: The AI overlord suggests `DNS hostname`

gleb-lobov: Anyways, the index is generated from the tool descriptions, and this is what it says there. If we want to fix it - I'll have to do a new release of snowplow-cli. Let me know if it's worth it

cksnp: No it's not worth it, I've already approved the PR, the word change was merely a suggestion

alexanderdean:
Hi @gleb-lobov & @cksnp - I think the updated docs are great.

My only suggestion at this point is that the Snowplow CLI is way too nested now in the docs nav for the current, let-alone future, scope of the Snowplow CLI.

The original scope before https://github.com/snowplow/snowplow-cli/issues/3 was very narrow - so the CLI docs reasonably lived under:

Event Studio > Programmatic management > Snowplow CLI

I now feel strongly that Snowplow CLI is a top-level construct - it doesn't nest under anything else. I would make the order in the left hand docs nav:

Fundamentals
Working with AI
Snowplow CLI
Event Studio

/cc @jpiekos for vis

jborlase-snowplow:
@alexanderdean - We should have a discussion on the role the Snowplow CLI plays, and what it should play going forward. In its current form it is used to enable customers to create, manage, and govern their tracking utilising the file system. The primary goal of this was to enable Git Ops for customers to govern their tracking. Hence its current location in the docs. I'm not sure adding the tracking-cli warrants creating a top-level object, for added context we see around 4-5 visitors per month looking at the tracking-cli docs, compared to 45-50 for the snowplow-cli. I add that to say, I don't believe adding the tracking-cli is that significantly impactful alone.

If I understand you correctly, how I believe you are intending the snowplow-cli to evolve into would require a change in the primary purpose the snowplow-cli. We are investing further into the MCP server to enable agentic management for a number of reasons - happy to chat through the rationale further.

gleb-lobov:
I don't think we have enough functionality to be placed right after the AI stuff. But I agree that current section is pretty deep.

What about moving snowplow-cli to be under the Components section? It seems like the appropriate place, and it's not as deep.

Let me know if it's appropriate - I'll move it. For now I'll merge it as is, since it's needed to unblock some planned agent improvements

alexanderdean:
Hi @gleb-lobov, @jborlase-snowplow - I am happy with moving to Components for now.

James - I agree that adding Tracking CLI is not very impactful alone (though of course number of visits to its docs is a function of its visibility not just its utility).

The direction of travel is pretty clear though - Snowplow CLI becomes the universal CLI for working with all Snowplow services. I am working on a couple of extensions to its capabilities now. This isn't coming from an ideological position - it's coming from a position of laziness on my part. I don't want to have to navigate all the Snowplow real estate from the last decade just to work with a Snowplow pipeline, I just want to use one tool.

It's a bit glib but agents are obviously the other large driver of this. See Cloudflare's very recent announcement of their own universal CLI:

https://blog.cloudflare.com/cf-cli-local-explorer/

gleb-lobov:
@jborlase-snowplow I've moved the cli section to components, can you check if you're happy with it's placement?

PR #1822, opened 2026-06-10
Agent classification enrichment

istreeter:
## AI reviews

Claude will automatically review this PR against the docs style guide.

If you have questions or want it to look again at something specific, tag `@claude` in a comment.

PR #1823, opened 2026-06-10
Enrich 6.12.0 including cloudfront/firehose adapter

istreeter:
- Bumps Enrich component version to 6.12.0
- Adds documentation for cloudfront adapter, because users can start using this immediately
- Documentation for agent classification enrichment is deferred to #1822 because it needs console support.

### DIFF
--- a/docs/sources/cdn-trackers/cloudfront/index.md
+++ b/docs/sources/cdn-trackers/cloudfront/index.md
@@ -1,0 +1,185 @@
+---
+title: "CloudFront access logs via Firehose"
+sidebar_position: 20
+sidebar_label: "CloudFront"
+description: "Track page views from CloudFront access logs by forwarding them to Snowplow through Amazon Data Firehose."
+keywords: ["cloudfront tracking", "cloudfront access logs", "amazon firehose", "cdn tracking", "ai agent tracking"]
+date: "2026-06-10"
+---
+
+If your website is served through Amazon CloudFront, you can forward access logs to Snowplow to capture page views from bots, AI agents, and other clients that don't execute JavaScript. Snowplow processes each log entry as a [page view event](/docs/fundamentals/events/index.md), which integrates naturally with your existing web analytics.
+
+This integration uses [CloudFront standard logging](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/AccessLogs.html) with Amazon Data Firehose as the delivery mechanism. Firehose forwards log batches to the Snowplow Collector via HTTP, where they are processed by the CloudFront Firehose adapter in Enrich. This adapter is available from Enrich 6.12.0.
+
+:::note[CDN tracking vs. web tracking]
+
+CDN-level tracking captures all requests, including from bots and agents that don't run JavaScript. However, for single-page applications, multiple client-side page views typically correspond to only a single CDN request. Treat CDN events and browser events as complementary rather than equivalent.
+
+:::
+
+## How it works
+
+CloudFront sends batches of access log records to a Firehose stream. Firehose forwards each batch to the Snowplow Collector, and Enrich produces one page view event per log entry.
+
+```mermaid
+flowchart LR
+  client(Client) --> cf(CloudFront)
+  cf -->|access logs| firehose(Firehose)
+  firehose -->|"HTTP POST /com.amazon.aws.cloudfront/firehose"| collector(Snowplow Collector)
+  collector --> enrich(Enrich)
+  enrich --> events(Page view events)
+```
+
+The following CloudFront log fields are mapped to Snowplow event fields:
+
+| CloudFront field | Snowplow field | Notes |
+|---|---|---|
+| `x-host-header` | `page_url` | Combined with `cs-uri-stem` and `cs-uri-query` to form the full URL |
+| `cs-uri-stem` | `page_urlpath` | |
+| `cs-uri-query` | `page_urlquery` | |
+| `cs(Referer)` | `page_referrer` | |
+| `cs(User-Agent)` | `useragent` | |
+| `timestamp(ms)` | `dvce_sent_tstamp` | Milliseconds since epoch |
+| `c-ip` | `user_ipaddress` | Optional — see note on IP collection below |
+
+In addition to the mapped fields, every event has the following fixed values:
+
+| Field | Value |
+|---|---|
+| `event` | `page_view` |
+| `platform` | `srv` |
+| `v_tracker` | `com.amazon.aws.cloudfront/firehose` |
+| `app_id` | `cloudfront` by default; override with the `aid` query parameter |
+
+To set a custom `app_id`, append `?aid=<value>` to the collector endpoint URL when configuring Firehose.
+
+:::note[IP address collection]
+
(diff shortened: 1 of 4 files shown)


docs/sources/cdn-trackers/cloudfront/index.md
```diff
@@ -0,0 +1,185 @@
...
+        result = classify(record)
+        output.append({
+            "recordId": record["recordId"],
+            "result": result,
+            "data": record["data"],
+        })
+    return {"records": output}
+```
+
+You can extend `AGENT_UA_SUBSTRINGS` to include additional user agents, or replace the filtering logic to suit your use case — for example, to also drop requests for static assets based on the `cs-uri-stem` field.
```
stanch: Could you add an example of URI filtering, same as what I did for Cloudflare?

benjben (approved):

docs/sources/cdn-trackers/cloudfront/index.md
stanch: @benjben

benjben: It's not easy to come up with URI filtering that would work for every website as they are all different.
I guess customers would need to fine-tune the filtering depending on their setup.

There are 2 approaches:
- Deny all and allow a whitelist.
- Allow all and deny a blacklist.

I went with the latter as I feel it is easier to determine what is a static asset that we don't want (e.g. css).

Please let me know if you don't agree.

---

### What gets kept

- HTML pages: `/about.html`, `/index.htm`
- Site root: `/`
- Dynamic backends: `/login.php`, `/Page.aspx`, `/checkout.jsp`, `/cart.do`
- Extension-less paths: `/about`, `/api/users`, `/blog/post-slug`
- Documents that act as content: `/whitepaper.pdf`, `/report.docx`, `/feed.xml`, `/manifest.json`, `/robots.txt` (debatable — these aren't in the asset list, so they pass)

### What gets dropped

- **Styles**: `.css`
- **Scripts**: `.js`, `.mjs`, `.map`
- **Images**: `.png`, `.jpg`, `.jpeg`, `.gif`, `.svg`, `.webp`, `.ico`, `.bmp`, `.tif`, `.tiff`, `.avif`, `.heic`
- **Fonts**: `.woff`, `.woff2`, `.ttf`, `.otf`, `.eot`
- **Audio/Video**: `.mp3`, `.wav`, `.ogg`, `.m4a`, `.flac`, `.mp4`, `.webm`, `.mov`, `.m4v`, `.avi`, `.mkv`, `.flv`
- **Archives**: `.zip`, `.tar`, `.gz`, `.tgz`, `.bz2`, `.7z`, `.rar`

### A few deliberate choices worth flagging

- **`.pdf`, `.json`, `.xml`, `.txt`** are NOT in the asset list. PDFs are often customer-facing content (downloads they care about); JSON/XML are often API responses. Customers who want these dropped can edit the tuple. The conservative default is "keep it" — false-positive page views are easier to deal with than missing real traffic.
- **`.map` is dropped**, since source-maps are dev tooling — but if a customer publishes user-facing tools that use `.map` extension for something else, they should remove it.
- **Case-insensitive matching** means `/IMG.JPG` is dropped just like `/img.jpg`.
- **Path-style edge cases** (`/path;jsessionid=…html`) match correctly via `endswith` against the lowercased stem.

If you want to be even more conservative (drop fewer things), the safest minimal set is `(".css", ".js", ".png", ".jpg", ".jpeg", ".gif", ".svg", ".ico", ".woff", ".woff2")`.

```
"""
Firehose data-transformation Lambda for the CloudFront ingestion adapter.

Keeps only access-log records that satisfy BOTH conditions:
  1. `cs(User-Agent)` looks like an AI agent (Claude, ChatGPT, etc.)
  2. `cs-uri-stem` is NOT an asset request (CSS, JS, images, fonts, media)

Page hits and any non-asset URL pass through: static HTML (.html / .htm), the
site root (/), dynamic backends (.php, .aspx, .jsp), REST/SPA routes
(extension-less, e.g. /about or /api/users). Only unambiguous static assets are
dropped.

Deployment:
- Runtime: Python 3.12 (or any 3.x supported by Lambda)
- Handler: firehose_filter_agent_traffic.lambda_handler
- Timeout: 60s is plenty (no I/O, just parse + match)
- Memory: 128 MB
- Attach to the Firehose stream's "Transform source records with AWS Lambda" step.
"""

import base64
import json
import logging

logger = logging.getLogger()
logger.setLevel(logging.INFO)

# Case-insensitive substrings used to identify AI-agent traffic in the User-Agent
# header. Extend as needed - each addition is a one-line tuple entry.
AGENT_UA_SUBSTRINGS = (
    "claude",        # Claude Code, Claude Desktop, anthropic-* clients
    "chatgpt",       # ChatGPT app
    "gptbot",        # OpenAI's crawler
    "openai",        # other OpenAI clients
    "gemini",        # Google Gemini
    "perplexity",    # Perplexity
    "copilot",       # GitHub Copilot / Microsoft Copilot
)

# Case-insensitive suffixes identifying static assets. URIs ending in any of
# these are considered asset fetches (CSS, JS, images, fonts, media, archives,
# source maps, favicons) and dropped. Everything else - including HTML pages,
# dynamic backends (.php, .aspx, .jsp), extension-less REST/SPA routes, and the
# site root `/` - passes through as a candidate page hit.
ASSET_EXTENSIONS = (
    # Styles
    ".css",
    # Scripts
    ".js", ".mjs", ".map",
    # Images
    ".png", ".jpg", ".jpeg", ".gif", ".svg", ".webp", ".ico", ".bmp",
    ".tif", ".tiff", ".avif", ".heic",
    # Fonts
    ".woff", ".woff2", ".ttf", ".otf", ".eot",
    # Audio / Video
    ".mp3", ".wav", ".ogg", ".m4a", ".flac",
    ".mp4", ".webm", ".mov", ".m4v", ".avi", ".mkv", ".flv",
    # Archives / downloads
    ".zip", ".tar", ".gz", ".tgz", ".bz2", ".7z", ".rar",
)

USER_AGENT_FIELD = "cs(User-Agent)"
URI_STEM_FIELD = "cs-uri-stem"


def is_agent(user_agent):
    """Return True iff `user_agent` looks like an AI agent."""
    if not user_agent or user_agent == "-":
        return False
    lower = user_agent.lower()
    return any(needle in lower for needle in AGENT_UA_SUBSTRINGS)


def is_asset_request(uri_stem):
    """Return True iff `uri_stem` ends with a known static-asset extension."""
    if not uri_stem or uri_stem == "-":
        return False
    return uri_stem.lower().endswith(ASSET_EXTENSIONS)


def classify(record):
    """Decide what to do with a single Firehose record."""
    try:
        payload = base64.b64decode(record["data"])
        log_entry = json.loads(payload)
    except Exception as e:  # malformed Firehose record - flag for inspection
        logger.warning("Could not parse record %s: %s", record.get("recordId"), e)
        return "ProcessingFailed"

    ua = log_entry.get(USER_AGENT_FIELD, "")
    stem = log_entry.get(URI_STEM_FIELD, "")
    if is_agent(ua) and not is_asset_request(stem):
        return "Ok"
    return "Dropped"


def lambda_handler(event, context):
    """Firehose passes a batch of records; we tag each as Ok / Dropped / ProcessingFailed."""
    output = []
    kept = dropped = failed = 0

    for record in event["records"]:
        result = classify(record)
        if result == "Ok":
            kept += 1
        elif result == "Dropped":
            dropped += 1
        else:
            failed += 1
        # Firehose requires `data` on every output record, even for Dropped/ProcessingFailed
        # (it is ignored in those cases). We never rewrite the bytes, so just echo them back.
        output.append({
            "recordId": record["recordId"],
            "result": result,
            "data": record["data"],
        })

    logger.info("processed=%d kept=%d dropped=%d failed=%d", len(output), kept, dropped, failed)
    return {"records": output}

```

PR #1827, opened 2026-06-11
Document how to obtain a Bearer token for Signals API auth

matus-tomlein:
## What changed?

Added an "Authenticate to the Signals API" subsection to the [Signals connection page](https://docs.snowplow.io/docs/signals/connection/), under the existing **Signals API** heading. It documents:

- That direct Signals API requests require a Bearer token in the `Authorization` header (the SDKs handle this automatically)
- That the token is a JWT obtained from the Console Credentials API by exchanging your Snowplow API key and key ID
- A minimal `curl` example for the token exchange and for using the JWT as a Bearer token in a Signals API request
- A note pointing Signals Sandbox users to their Sandbox token

It links to the [account management](https://docs.snowplow.io/docs/account-management/) page for the full token-exchange process rather than duplicating it.

## Why?

Customer feedback (AISP-1378): figuring out how to obtain an auth token for the Signals API required digging into the Python SDK source. This makes the Bearer token flow explicit in the docs.

Closes AISP-1378.

## Reviewer guidance

The example Signals request path uses `{{API_URL}}/api/v1/...`, matching the `api/v1/` prefix referenced elsewhere in the Signals docs. Worth confirming that's an appropriate example endpoint.

## AI reviews

Claude will automatically review this PR against the docs style guide.

If you have questions or want it to look again at something specific, tag `@claude` in a comment.


stanch (changes requested):
This documentation already exists here, unless I am missing something: https://docs.snowplow.io/docs/account-management/. Could we reuse it or link to it? (I admit the naming of that section is a bit confusing.)

stanch:
Oh, you are already linking there 🤦

matus-tomlein:
@stanch would it be better to remove the instructions here and only keep the link? thought it would be easier for users to follow without having to click through, but can remove the duplication if it's preferable to have the instructions just in one place.

jborlase-snowplow (approved):
LGTM

stanch (approved):

PR #1828, opened 2026-06-11
Restructure Signals docs along the define/consume seam

jborlase-snowplow:
## What this does

Restructures the Signals docs section so that **defining configurations** (data team workflow) and **consuming values in applications** (app developer workflow) are clearly separated. This resolves three navigation/clarity complaints plus a set of issues found in a full section review. A second pass removes the Signals Sandbox (retired in favor of the Snowplow free trial) and splits setup from connection.

### New top-level structure

```
Signals
├── Introduction
├── Get started
├── Fundamentals
├── Set up Signals                   ← new (Console enable flow, split from Connect)
├── Define attributes
│   ├── Attribute groups / Attributes / Attribute keys / Warehouse configuration
├── Define interventions
├── Use in your application          ← new
│   ├── Services                     ← moved from Attributes
│   ├── Retrieve attributes          ← new (split out of the attributes hub)
│   └── Subscribe to interventions   ← moved from Interventions
└── Connect to Signals               ← now SDK/API reference, moved to the end
```

### Key changes

- **"Attributes" → "Define attributes"**: kills the `Attributes > Attributes` sidebar duplication. The hub index is now a genuinely simple landing page, and the shared publish/unpublish/delete lifecycle section moves here from the connection page.
- **New "Use in your application" category**: retrieval reference content is no longer hidden inside a category landing page. Services move here since they exist to serve applications.
- **Setup/connect split**: the one-time Console enable flow gets its own prominent "Set up Signals" page after Fundamentals (Console-first experience); "Connect to Signals" keeps its URL but becomes pure SDK/API reference (credentials, clients, method tables) at the end of the section, opening with the two reasons to connect and a capability table per interface.
- **Signals Sandbox removed**: all Sandbox references in the docs section replaced with the [Snowplow free trial](https://snowplow.io/get-started/snowplow-free-trial). This removes the Sandbox tabs from the connection page (every example is now a single code block), the Sandbox option on the Snowplow Inspector integration page, and the Sandbox screenshot. **Tutorials still using the Sandbox are deliberately untouched** — they'll be reworked in a follow-up once the free trial flow is live; only their deep links to removed anchors were retargeted.
- **Interventions Python SDK page merged** into the main interventions page as Console/Python SDK tabs, matching the pattern used by every attributes page.
- **Hidden SEO page** (`signals/index.md`) trimmed to a short summary — it had already drifted from the get-started page it was copied from.
- **Redirects**: 301s added in `worker/redirects.js` and `static/_redirects` for the moved/removed URLs; the existing `receive-interventions` rule updated to avoid a redirect chain. All internal links in docs and tutorials updated, including renamed heading anchors.

### Mechanical fixes folded in

- Broken relative link on the interventions page (missing leading `/`)
- Entity property sum example compared a product `price` to the string `"transaction"` — now filters on the ecommerce action `type`
- Invalid UUID in a subscribe example; `Profile Store` → `Profiles Store`; "Event Collector" links pointed at the fundamentals index instead of the Collector page
- Orphaned paragraph above the attribute-types table on the fundamentals page
- Gerund/noun headings changed to imperative per the style guide; "more than one criteria" grammar
- Introduction cards relabeled and retargeted (e.g. "Define attributes" no longer points at attribute groups)

### Flagged, not fixed

The structured-events vendor string is inconsistent between the Console tab (`com.google.analytics.measurement-protocol`) and the SDK tab (`com.google.analytics`) on the Define attributes page. I couldn't verify what the Console dropdown actually shows, so I left both as-is — worth a check from someone with Console access.

## Verification

Full `yarn build` passes after each change set with `onBrokenLinks: 'throw'` and `onBrokenAnchors: 'throw'`, so every internal link and anchor in the site is validated against the new structure.


docs/signals/interventions/index.md
matus-tomlein, line 294: Maybe somewhere in this file (and similar in the define attributes page), we should link to the page with subscribing for interventions? This section requires having a subscription to make sense. It also is the natural next question after defining an attribute group/intervention – how do I retrieve the values?

docs/signals/applications/services/index.md
```diff
@@ -1,7 +1,7 @@
 ---
```
matus-tomlein, line 1: I can see the thinking behind putting this into the "Use in your application" section, but still seems like it breaks the flow a little bit, since the page is about defining services which is not what you do in your application, but you more likely do it alongside with defining the attributes? Shouldn't the be a top-level page like "Define attributes"?

docs/signals/interventions/index.md
jborlase-snowplow: Good call — added in 8d3a96e4: subscription links from the intervention publish section, the test section (with a note that the test only arrives if you're already subscribed to that attribute key ID), and direct interventions. Also added a pointer after publishing attribute groups to retrieving the values via a service.

docs/signals/applications/services/index.md
jborlase-snowplow: We went back and forth on this one. Keeping it here deliberately: services exist to provide a stable consumption interface for applications (that's how the fundamentals page frames them too), and the Retrieve attributes page leans on them, so they sit next to the consumption docs rather than buried under attributes. You're right that defining one happens alongside defining attributes though — the define-side pages now cross-link to services at the point you'd need them (versioning, publish next-steps). Happy to revisit if it trips people up in practice.

jborlase-snowplow: ^ that was Claude 😂

matus-tomlein (approved):
LGTM!

PR #1829, opened 2026-06-15
Add identifier selection guidance and scale limits to Identities docs

jborlase-snowplow:
## What

Adds guidance to help customers avoid configuring Identities on unstable, high-cardinality identifiers, which inflates the identity graph database and drives up cost and latency. Also documents that Identities is self-serviceable only within supported scale.

## Changes

- **New page** [`choosing-identifiers/index.md`](docs/identities/choosing-identifiers/index.md):
  - What makes a good (stable) vs unstable identifier, and why each distinct identifier value becomes a node in the graph database.
  - Why high-cardinality identifiers increase cost, increase latency, and degrade resolution quality.
  - Warehouse validation to run **before** configuring: a cardinality check and an "identifiers per unique identifier" fan-out check, with SQL templates and an interpretation table.
- **`configuration/index.md`**: new `Supported scale` section stating the deployment is self-serviceable for typical volumes and reasonable identifier configurations, and directing high-volume deployments to Snowplow Support (qualitative, no published threshold). Added a tip linking to the new page.
- **`index.md`**: Get started now points to choosing identifiers as the first step and notes self-serve within supported scale.
- Renumbered sidebar so Choosing identifiers sits between Concepts and Configuration.

## Validation

- `yarn build` passes with no broken links or warnings.

### DIFF
--- a/docs/identities/choosing-identifiers/index.md
+++ b/docs/identities/choosing-identifiers/index.md
@@ -1,0 +1,85 @@
+---
+title: "Choosing identifiers for Identities"
+sidebar_label: "Choosing identifiers"
+sidebar_position: 2
+date: "2026-06-15"
+description: "How to choose stable identifiers for Snowplow Identities and validate their cardinality against your warehouse before configuring them."
+keywords: ["identities", "identity resolution", "identifiers", "cardinality", "best practices"]
+---
+
+The quality of identity resolution, and the cost and latency of running Identities, both depend on which event fields you configure as [identifiers](/docs/identities/concepts/index.md#identifiers). Before adding an identifier in [Console](/docs/identities/configuration/index.md), choose fields that are stable for each user, and validate their cardinality against your warehouse.
+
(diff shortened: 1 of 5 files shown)


jborlase-snowplow:
Addressed the style review in 46d67ff5:

- **List capitalization** (material): capitalized the first word of each prose list item across the three lists. Items led by a code identifier (`user_id`, `domain_userid`) left as-is, per the suggestion.
- **Single-digit numbers** (minor): table now reads "Around one" and "two to five".
- **Support link**: confirmed valid (403 is bot protection), no change.

Also caught two more while re-checking the style guide:
- Removed the banned word "currently" from the new deployment-support note.
- Matched on-page badge naming ("Cloud", "Private Managed Cloud").

Separately, the "Choosing identifiers" page is now nested under Configuration per review feedback. `yarn build` passes with no broken links.

jamessnowplow (approved):
Solid advice.

PR #1836, opened 2026-06-19
An end-to-end tutorial to demo signals w/ python

cksnp:
## What changed?

Added a tutorial for e2e signals implementation w/ python for tracking & signals control plane.

## Why?

Tutorial value aside, this is very useful for the snowplow assistant.

### DIFF
--- a/tutorials/python-tracking-and-signals/define-attributes.md
+++ b/tutorials/python-tracking-and-signals/define-attributes.md
@@ -43,13 +43,13 @@ sp_signals = Signals(
 
 ## Define an attribute group
 
-An attribute group computes one or more attributes for a given [attribute key](/docs/signals/concepts/) (the identifier the attributes are calculated against). Here you'll key on `domain_userid`, and count `task_completed` events over a rolling 7-day window.
+An attribute group computes one or more attributes for a given [attribute key](/docs/signals/concepts/) (the identifier the attributes are calculated against). Here you'll key on `user_id`, and count `task_completed` events over a rolling 7-day window.
 
 The `Event` you reference must match the schema you created: same vendor, name, and version.
 
 ```python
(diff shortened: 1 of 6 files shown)


tutorials/python-tracking-and-signals/set-up-tracking.md
```diff
@@ -0,0 +1,152 @@
...
+
+# A reusable project entity, attached to events as context.
+project = SelfDescribingJson(
+    "iglu:com.example/project/jsonschema/1-0-0",
+    {"project_id": "proj_42", "plan": "team"},
+)
+
+# A page view (for web surfaces) and a screen view (for app surfaces).
+tracker.track(PageView(page_url="https://app.example.com/board", page_title="Board"))
+tracker.track(ScreenView(id_=str(uuid.uuid4()), name="Project board", context=[project]))
```
mfolcini, line 103: isn't this a mobile schema and this is a BE project?

tutorials/python-tracking-and-signals/define-attributes.md
```diff
@@ -0,0 +1,155 @@
...
+                major_version=1,
+                path="priority",
+            ),
+            period=timedelta(days=7),
+        ),
+    ],
+)
+```
+
+`domain_userid` is a built-in attribute key exported by the SDK. `domain_sessionid` and `user_id` are also available if you want to compute attributes per session or per logged-in user instead.
```
mfolcini: why are we choosing domain_userid vs user_id as core identifier if the user is identified?

cksnp: Good point, will do.

Reminder for me: also check if conclusion (how to get domain ID) is still relevant after that

mfolcini (approved):

PR #1869, opened 2026-07-23
[AISP-1508] Document the "changed" intervention operator

matus-tomlein:
## Summary

Documents the new **`changed`** intervention criterion operator on the [Define interventions](https://docs.snowplow.io/docs/signals/interventions/) page. `changed` triggers an intervention every time a referenced attribute's value changes, rather than only the first time a condition becomes true.

Completes epic AISP-1504 (API: AISP-1503, Console: AISP-1506, SDK: AISP-1507 — this is the docs).

## Changes

`docs/signals/interventions/index.md`:
- Added a `changed` row to the operators table (no value required).
- Clarified the "Sent once" note to call out `changed` as the exception (it sends on every change).
- Added a **The `changed` operator** section covering:
  - triggers on every value change, including the first set (null → value);
  - requires Signals **0.14.6+**;
  - only supported on its own or within an `all` group (not `any`/`none`);
  - change detection is best-effort under high-concurrency same-key bursts;
  - how it behaves in the preview (treated as always matching, so other criteria still filter).

## Testing

`npm run build` passes. With `onBrokenLinks: 'throw'` and `onBrokenAnchors: 'throw'`, this confirms the two new in-page anchors (`#the-changed-operator`, `#preview`) resolve. Prettier clean.

### DIFF
--- a/docs/signals/interventions/index.md
+++ b/docs/signals/interventions/index.md
@@ -77,7 +77,7 @@ The table below lists all available arguments for a `RuleIntervention`:
 Criteria are the conditional rules that determine when an intervention should trigger.
 
 :::note[Sent once]
-An intervention is sent only the first time the criteria are met. Read an example of how this works in the [delivery example](/docs/signals/applications/subscribe/index.md#delivery-example).
+An intervention is sent only the first time the criteria are met. Read an example of how this works in the [delivery example](/docs/signals/applications/subscribe/index.md#delivery-example). The [`changed` operator](#the-changed-operator) is the exception: it sends the intervention every time the attribute's value changes.
 :::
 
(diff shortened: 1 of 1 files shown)


docs/signals/interventions/index.md
```diff
@@ -158,6 +159,22 @@ criteria = InterventionCriteriaAll(all=[
...
 
+### The `changed` operator
+
+Most operators trigger an intervention the first time the criteria become true. The `changed` operator is different: it triggers every time the attribute's value changes, regardless of the new value. Use it to react to updates to an attribute, for example to re-fetch a user's profile whenever any of their attributes change.
+
+`changed` takes no value. It triggers whenever the attribute's previous and current values differ, including when the attribute is set for the first time (from no value to its first value).
+
+:::note[Requires Signals 0.14.6 or later]
+The `changed` operator is available from Signals version 0.14.6.
+:::
```
igneel64: I am not sure it has some value mentioning the version since all have been updated.

igneel64 (approved):

PR #1881, opened 2026-07-28
Document event specification validation for Micro development environments and good/bad routing

patofp:
**DPS-2329**: https://snplow.atlassian.net/browse/DPS-2329

- Development environments now have a section documenting that they validate incoming events against every published version of your event specifications and against the latest draft of each one, so tracking can be tested without publishing the tracking plan or its specifications.
- The validation and inference pages point to development environments as the way to test drafts and inference without affecting production.
- The inference page's list of attributes that matching ignores now names source applications explicitly.
- The validation page gains a "Send invalid events to failed events" section describing the per-tracking-plan routing setting (with a screenshot of the Data quality rules control), and its default-behavior note and warehouse cases are now conditional on that setting.


patofp:
@claude please run the style review again. The sections about development environments and inference were reworded since your first pass, and the branch is rebased on main.

patofp:
Thanks, applied most of this.

Fixed: the ambiguity between the declared version and the draft, where the setting lives in Console and the permission it needs, the inference wording on both pages, the Draft status bullet (a pipeline does not match drafts, a development environment does), the repeated default, the 60 word sentence, the missing period on the list item, "changes take effect within a few minutes", plainer wording for the fallback case, splitting the two unrelated ideas in the Micro section, "test" instead of "try", "using" instead of "following", the image filename, and moving the new paragraph above the example.

Two I skipped.

Release note: both capabilities are already released and have been running in production for a while, so this PR documents what shipped rather than announcing it. Updating the May note is worth doing, but as its own change.

Frontmatter of the Micro in Console page: out of scope here. The missing date field predates this PR.

@claude please review again with these changes in.

patofp:
Applied, thanks.

Added the release note: `release-notes/test-event-specifications-and-route-validation-failures/index.md`, covering both capabilities and linking back to the docs pages.

Also applied: the navigation entry point and the **Edit** permission on tracking plans, the default before the alternative, active voice in the admonition and in the routing sentence, the missing specification case moved into "Validation entity" with only its routing consequence left in the configuration section, backticks on the warehouse `events` table, the clearer back reference for the three warehouse cases, the alt text naming both options, the Draft bullet keeping the article style of its siblings, the link to the development environments section instead of re-explaining the draft rule, the split of the 42 word sentence, the singular and plural consistency, "the draft" instead of "the current draft", and dropping the redundant half about publishing the tracking plan.

Two things for the next pass. Please don't contradict the previous review: the earlier one asked for the new inference paragraph to go above the **Example** heading, this one asked for the end of the section next to the tip, and I followed this one. And please don't ask for changes outside the scope of this PR, such as the missing `date` field in the Micro in Console frontmatter, which predates it.

@claude please review again with these changes in.

patofp:
Applied the ones that belong to this change: the Draft bullet rewrite, the link on source applications, the reordered final paragraph in the routing section, the permission anchor, "in the Console", the simpler step 3, and in the release note the active voice, the entity names in backticks and the link to the Micro dashboard.

Not applying the rest, all of it outside this change: the `date` field and the italic permissions on the Micro in Console page, the frontmatter of the validation page, a new sentence about which version pipeline inference uses, and the addition to the failed events page. That last one is a fair point and worth its own PR.

I'm stopping the automated passes here. Each round has reopened items I had already declined as out of scope and added new ones on pages this PR doesn't touch, so the suggestions are growing rather than converging. The content is verified against the pipeline and Console code, the style points raised across the four passes are in, and what the PR needs now is a human reviewer.

docs/event-studio/tracking-plans/event-specification-validation/index.md
```diff
@@ -13,20 +13,22 @@ Event specification validation checks whether incoming events conform to the rul
...
 - **Passed validation, or matched by inference**: the event has an `event_specification` entity but no `event_specification_validation` entity. Either the event was Snowtype-tracked and passed validation, or the pipeline matched it to a specification by inference.
 - **Not associated with a specification**: the event has neither entity. No specification was attached at tracking time, and the pipeline did not match the event to a published specification.
 
+These cases assume the default **Data quality rules** setting. If the tracking plan [sends invalid events to failed events](#send-invalid-events-to-failed-events), events that fail validation don't reach your warehouse `events` table.
+
 ## Validation entity
 
-The pipeline attaches an `event_specification_validation` entity to events that fail validation.
+The pipeline attaches an `event_specification_validation` entity to events that fail validation. It also attaches one when it cannot find the declared specification, either because it was never published or because its rules are invalid.
```
gleb-lobov: > or because its rules are invalid.

What does it mean here?

patofp: It means event spec instructions or rules, I'll try to make it more clear

gleb-lobov (approved):

PR #1890, opened 2026-07-29
Add tutorial: personalize by account with custom Signals attribute keys

jborlase-snowplow:
## What changed?

New tutorial `tutorials/signals-account-attributes/` (6 pages + 2 live Console screenshots): track a B2B account entity, define a custom `account_id` attribute key (Console and SDK tabs), compute cross-user account attributes, and retrieve them with a rule intervention.

## Why?

Fills the account-level/custom-attribute-key gap in Signals tutorial coverage (tutorials gap analysis). Existing tutorials only cover the built-in session/user keys.

## Reviewer guidance

- Every code block was executed in page order from a fresh venv against a trial org; attribute values and the intervention were asserted via the Signals API with exact expected values. Console create-key flow verified live; screenshots captured from it.
- The UUID caveat was tested explicitly: non-UUID account IDs DO compute/retrieve attributes; only intervention subscription rejects them. The sibling `python-tracking-and-signals` tutorial overstates this — fix proposed separately, not in this PR.
- The conclusion links `/docs/signals/attributes/warehouse-config/` instead of the (unbuilt) warehouse-attributes tutorial; swap when that tutorial lands.
- Found during verification, flagged for separate follow-up: Console create-key form has no name field (attribute-keys docs outdated); groups 422 if published before their custom key; ~1–2 min engine config-apply lag contradicts "as soon as published" docs wording.


jborlase-snowplow:
Pushed two commits addressing both review tracks:

**`b7fa4d14` — newcomer-review fixes** (an independent fresh-eyes review of the tutorial; findings ranged from a wrong Python prerequisite to silent-failure traps):
- Prerequisites corrected: Python **3.11+** (every `snowplow-signals` release requires it), Console permissions incl. the Admin-only production migration
- The data-structure step now spells out draft → development → production, and a new **"Confirm your events arrived"** checkpoint (with a real `EmitterConfiguration(on_failure=...)` example, demonstrated against a deliberate bad host) closes the tutorial's worst failure mode: silent event loss misread as Signals latency
- The Console tab path no longer dead-ends (the `AttributeKey` object is restated inline); `account_id` is printed and the single-Python-session assumption is stated; the final script's climax is guarded instead of ending in an uncaught `queue.Empty`; a cleanup section was added
- All code re-executed end-to-end against a live org; the full expected-output block is a genuine capture

**`6ae7894a` — the 14 style-review items** — all applied (admonition types, list conventions audited script-wide, H2 counts, keywords, etc.). One deliberate exception: the draft-data-structure warning stays a `:::warning` because stopping at draft genuinely loses events, which is what the type is reserved for.

Two review findings were confirmed as bugs in *other* docs pages and deliberately not fixed here: the attribute-keys page's Console tab (name/owner fields that don't exist) and the interventions page typing `target_attribute_keys` as `string[]`.

jborlase-snowplow:
Pushed `9516833b` + `807f2ea2` — editorial sweep per product direction: verification-run artifacts are out of the reader-facing pages. Error strings and status codes across all troubleshooting sections are restated as positive instructions or checks; measured wait durations are neutral phrasing; the deliberate-failure exercise keeps its teaching value without the run transcript. The published-group immutability constraint is retained as a single positive sentence, phrased consistently with the interventions tutorial and the agentic-contexts docs.

jborlase-snowplow:
Pushed `664fbffa` — product-owner edit round (Console/Assistant-first, less process ceremony):

- Removed: the time-estimate and Console-permissions paragraphs, the Python-version failure explanation (the 3.11+ prerequisite stays), the draft→development→production walkthrough and admin-only warnings (now simply: both data structures need to be in production), the catalog/propagation panels, "When built-in keys are enough", and "Common connection mistakes"
- Event verification is now a positive step — check Console metrics, or ask the **Snowplow Assistant** to verify events arrived / check failed events (copy-paste prompt included)
- **Snowplow Assistant prompts added as the suggested route** for creating the data structures and — via new Console-default tabs matching the key section — the attribute group, service, and publish steps. Every prompt's intent was verified against the live API (field-by-field match with the Python definitions; publish exercised; teardown clean)
- Two verification notes: the deployed `account` schema has `account_id` + `plan` (no `name`), and prompts match it exactly; `owner` isn't settable via the Assistant surface — Console defaults it to your user, which the page now states
- No dependency on the Python tracking tutorial remains; both sibling references removed

igneel64:
A few thoughts, not saying all are actionable: 

- It overlaps a lot with python-tracking-and-signals. Same app, some same schemas and copy pasted paragraphs. The new thing here is really the custom attribute key, but the tutorial rebuilds all the tracking and Signals plumbing around it and ends up a good bit longer than the one it's repeating. Could this be a follow-on instead or linking back ? E.g. "same as the Python tutorial, but with an account entity instead of project".

- `payload_attribute_groups` This attribute is not documented anywhere though it is available on our SDK. Should be added on the model RuleIntervention table.

The Console/Python tabs. Three pages say to pick whichever suits you, then the retrieve-and-intervene page is Python only, so the Console reader has to switch anyway. They also still have to set OWNER, which the same page tells them they don't need. And with Console-vs-SDK plus Assistant-vs-by-hand inside Console, there's a fair bit to decide for something meant to be one happy path. Maybe just pick one?

- The tutorial says there's no name field and the screenshot proves it, but attribute-keys/index.md still says the form needs a unique name. Worth fixing maybe in this PR (?)

- The reason for using domain_userid is deferred to a page two steps later. Probably easier to just say it there.

- `Use a UUID for the account ID` May this should be just a link on interventions ? The info should be there. _Or we can even put this in the attribute keys page._

PR #1892, opened 2026-07-29
Cross-link agentic contexts from personalize-travel agent tutorial

jborlase-snowplow:
## What changed?

One note in `tutorials/signals-personalize-travel/personalizing-agent.md` pointing readers to the agentic-contexts docs and the Vercel AI SDK tutorial for the native pattern, framing attributes as the user-level profile complement. Date bump; nothing else touched.

## Why?

The travel demo app can't adopt agentic contexts today: its source is private (`snowplow-incubator/signals-demo-travel-planner`), the published image is ~4 months stale with no CI, and it pins `@snowplow/signals` SDKs at 0.2.0 (the feature ships in core 0.4.0). Scoping assessment concluded: cross-link rather than retrofit. Full app-side change list (8 items, ~1 dev-day + coordination) is documented if the owning team picks it up.

## Reviewer guidance

- **Merge-order dependency:** the note says "to see the native pattern, follow the Vercel tutorial", which is accurate once `update/vercel-agentic-context` merges (retrofit in progress). Merge this at or after that PR, or soften the sentence.
- Suggested follow-up ticket regardless of adoption: republish the demo image from source HEAD (it's 4 months behind and includes a `getSignals` URL fix), covering both this and the accelerator tutorial.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1884, opened 2026-07-28
Add free trial note to Signals tutorial intros

jborlase-snowplow:
> This PR was generated automatically by Claude.

## What changed?

Only 1 of the 8 Signals tutorials mentioned the Snowplow free trial, even though the trial includes Signals. This PR adds an access note to the intro page of the other 7 tutorials, modeled on the one in `tutorials/python-tracking-and-signals/introduction.md`. Each note is worded to fit the page and ends with the same sentence pointing to the [Snowplow free trial](https://snowplow.io/get-started/snowplow-free-trial).

While editing the intros, "get access" wording was retargeted from `/docs/signals/connection/` (the credentials page) to `/docs/signals/setup/` (the enable-Signals page):

- `signals-quickstart/start.md`: "set up a Signals connection" is now "enable Signals", the "Connecting to Signals" heading is now "Enable Signals", and the "A Signals connection" prerequisite is now "Signals enabled on your Snowplow account"
- `signals-ai-agent-context/introduction.md` and `signals-google-adk-agent/introduction.md`: "Signals deployed" links now point to the setup page as "Signals enabled"
- `signals-ml-prospect-scoring/intro.md`: the "Connect to Signals" prerequisite step now says "Enable Signals" and links to the setup page
- `signals-agentic-accelerator/intro.md` and `signals-personalize-travel/intro.md`: existing "Signals enabled" prerequisites now link to the setup page

## Why?

The free trial includes Signals, but readers landing on these tutorials without a Snowplow account had no pointer to it. The connection page links sent readers to a credentials page when they actually needed to enable Signals first.

## AI reviews

Claude will automatically review this PR against the docs style guide.

If you have questions or want it to look again at something specific, tag `@claude` in a comment.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1896, opened 2026-07-30
Pair agentic contexts with profile attributes in the Vercel AI tutorial

jborlase-snowplow:
## What changed?

Retrofits `tutorials/signals-ai-agent-context/` to use agentic contexts natively: an agentic-context definition step (Console + Python SDK tabs), and `lib/signals-context.ts` now fetches BOTH the profile attributes (service) and the in-session narrative (`getAgenticContext`, `format: "narrative"`) via `Promise.allSettled`, landing both in the system prompt with when-to-use-which guidance. Replaces the hand-rolled `formatAttributes()` block for session activity while keeping attributes as the computed-profile complement (per product direction). Also: fixes a real guard bug (a fresh session returns every attribute as `null` — the object is truthy, so the old empty-check never fired and `page_views_count: null` reached the prompt), and replaces the stale architecture PNG with a Mermaid diagram covering both fetches.

## Why?

The flagship agent tutorial pre-dated agentic contexts and hand-rolled what the feature now provides.

## Reviewer guidance

- Verified on a live Next.js scaffold up to the model call: both fetches execute, the composed prompt contains both sections, and the fixed guard was proven side-by-side against real API responses (old guard leaks nulls, new one doesn't). The model call itself is unverified — the tutorial's provider key (`AI_GATEWAY_API_KEY`) wasn't available; no provider was substituted.
- **In flight:** a browser pass is adding the two Console screenshots and verifying the Basic Web template's attribute list — those commits land on this PR shortly.
- Docs error found while verifying: `format: "narrative"` output leads with the configured prompt line before `[START CONTEXT]`; the applications docs page omits it — needs its own fix.
- Merge note: #1884 touches this tutorial's introduction too (trivial conflict); merging #1884 first is simplest. The personalize-travel cross-link PR (#1892) should merge at/after this one.


jborlase-snowplow:
Pushed `ee64aa81` — the Console verification pass, which closes the last content gaps:

- **Two live Console screenshots** added to the definition step, captured on the reader-facing `web_agent_activity` resource
- **Reader-blocking bug fixed:** the tutorial's pre-existing service name `web-agent-context` is rejected by the product — Signals names allow letters, numbers, and underscores only. Renamed to `web_agent_context` (now consistent with the ADK tutorial) with the constraint stated on the page
- **Run Preview claims removed** — verified absent across the create/draft/published/edit lifecycle on a trial org (likely warehouse-gated; the unconditional instruction was wrong for the tutorial's stated trial-pipeline audience)
- **Basic Web template verified live:** all four asserted attribute names correct; `unique_pages_viewed` confirmed as full URLs (`unique_list` over atomic `page_url`), stated explicitly and contrasted with the agentic context's `page_urlpath`

Remaining known gap (unchanged): the model call itself needs `AI_GATEWAY_API_KEY`; everything up to and including the composed system prompt is verified.

jborlase-snowplow:
Pushed `9a340de8` — editorial sweep per product direction: the SDK-behavior warning is replaced by the positive Console edit-and-publish instruction, and the null-handling section teaches the defensive pattern factually without bug framing.

jborlase-snowplow:
Pushed `e69dbaf7` — owner-lens edit round: the intro's Micro/full-pipeline panel removed (prerequisite now "A Snowplow account and pipeline with Signals enabled"), a new **Ask the Snowplow Assistant** section leads the Signals configuration page (copy/paste prompt, MCP-verified against the tutorial's definitions), rationale-after-example prose and duplicated instructions cut, troubleshooting reframed positively, and dates/titles tidied. Load-bearing verified constraints (the `page_url` distinction, publish-before-service ordering, the null-handling walkthrough) were deliberately retained as positive prose. Merge-order note: #1884 edits the same intro lines — whichever merges second should keep one prerequisite bullet and no panel.

igneel64:
Nothing here is blocking, just things I think are worth considering.

- Diagram. It's probably more detail than the reader needs. Also I would question the: Vercel AI Gateway → ChatWidget. It suggests the gateway streams to the browser, but with streamText() and toUIMessageStreamResponse() it streams to the /api/chat route, which then streams on to the browser. Maybe flatten it to browser, pipeline, Signals, route, gateway, and send the response back through the route.

- "designed for npm-based frameworks" isn't really a thing. Could we say "the tracker is distributed as npm packages you import into your components, rather than loading a tag on the page" ?

- Session ID. The tutorial reads getDomainUserInfo() and indexes [6]. There's a getDomainSessionId() method that does this directly (docs: getting-cookie-values/#domain-session-id), and our agentic contexts page already points people at it. If we use that, the paragraph about the _sp_id cookie and the array index can go too.

- "The scaffolded app/layout.tsx already includes font imports, a lang attribute, and CSS classes." Maybe drop this? A line saying to add SnowplowTracker to the root layout seems like enough.

- Configure Signals page. Bit odd that the attribute group is Console-only but the agentic context gets Console and Python tabs. Either add tabs everywhere or make it all Console. I'd lean towards Console-only, with a pointer to /docs/signals/agentic-contexts/ for the SDK, since the 45-line EventLog sample duplicates what's already there.

- `## Recent session activity (Snowplow Signals agentic context)`. The agentic context in narrative format already comes with its own prompt and heading, so we're probably adding a heading on top of a heading.

- "Note the null filter in getProfileSection()…" and the all-nulls JSON code block. This repeats the code comment right above it in the sample, so maybe both can go.

- The paragraph about the agentic context's prompt. It currently sits after createAgent(), a fair way from the narrative it's describing. Could move it straight after "The opening summary and the event table are generated by Signals…", and I don't think the "as well as from BASE_INSTRUCTIONS" bit adds much.

- Also if you think it will make the tutorial simpler we can skip the service creation, _fetching an attribute group is already supported_, and then introduce it at the end saying "if you want to combine, you can look at services'

PR #1908, opened 2026-08-13
Update warehouse docs

agnes-kiss:
## What changed?

Refined warehouse / batch engine + SDK related signals docs as per the upcoming release (and anything recent that we have not yet updated).


jborlase-snowplow (approved):
LGTM!

